Issue #24, opened 2017-12-28
Shell windows created for each node on Windows


afragapane:
### Description

Title.

### Steps to Reproduce

1. start the cli on windows using cygwin
2. run command: create net wintest
3. observe shell windows created for each node (node.exe)

**Expected behavior:**  Creating a node should not create a window

**Actual behavior:** Creating a node creates a window

**Reproduces how often:** Every time

### Versions
v0.0.1

afragapane:
Fixed by allowing Nodes to run as detached.

dicarlo2:
#34

PR #34, opened 2017-12-31
Nodes no longer create pop up windows on Windows OS -windowsHide: true

afragapane:
### Requirements

### Description of the Change
Changed fix from detached: true to windowsHide: true


dicarlo2:
Closes #24

Issue #20, opened 2017-12-26
delete command should prompt for confirmation


dicarlo2:
### Description

`neo-one delete` (or aliases) should prompt to confirm.

### Steps to Reproduce

1. `create net neo`
2. `delete net neo`

**Expected behavior:** Prompts to confirm deletion

**Actual behavior:** Deletion happens immediately

**Reproduces how often:** 100%

### Versions

v0.0.1

### Additional Information

Code Pointers:
- Deletion command is defined [here](https://github.com/neo-one-suite/neo-one/blob/90e537ee7d1b49da0986a2b9054783a240eac50f/packages/neo-one-cli/src/interactive/createCRUD.js#L326)

afragapane:
Taking a look at this

dicarlo2:
Looks good. I'll probably open another issue later for prompting about the dependent resources that will be deleted.

PR #62, opened 2018-02-02
Delete prompt now works as argument from cli

afragapane:
### Requirements

### Description of the Change
Previously, delete prompt only worked in interactive cli and hung when delete command passed as an argument (e.g. neo-one delete net altest).  Now the prompt works the same in both cases.

### Test Plan
Tested delete command and all prompt options in both interactive and as an argument.

### Benefits
Prompts in both use cases

### Applicable Issues
#20

### DIFF
--- a/packages/neo-one-cli/src/interactive/createCRUD.js
+++ b/packages/neo-one-cli/src/interactive/createCRUD.js
@@ -11,6 +11,7 @@ import {
   type GetCRUD,
   type Plugin,
   type TaskStatus,
+  type InteractiveCLI,
   getTasksError,
   areTasksDone,
 } from '@neo-one/server-plugin';
@@ -26,8 +27,6 @@ import stripAnsi from 'strip-ansi';
 import { empty } from 'rxjs/observable/empty';
 import logUpdate from 'log-update';
 import { timer } from 'rxjs/observable/timer';
-
-import type InteractiveCLI from '../InteractiveCLI';
 
 const pointer = chalk.yellow(figures.pointer);
 const skipped = chalk.yellow(figures.arrowDown);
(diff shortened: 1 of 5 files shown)


packages/neo-one-cli/src/interactive/createCRUD.js
```diff
@@ -151,22 +153,24 @@ const renderTasks = (
...
-|}) =>
-  cli.vorpal.activeCommand.prompt({
-    type: 'confirm',
-    name: 'continue',
-    default: false,
-    message: `Are you sure you want to delete ${
-      crud.resourceType.name
-    } ${name}?`,
-  });
+|}) => {
```
dicarlo2: nit: I'd leave it the way it was before, just replace `cli.vorpal.activeCommand.prompt` with `inquirer.prompt`

packages/neo-one-cli/src/interactive/createCRUD.js
```diff
@@ -151,22 +153,24 @@ const renderTasks = (
...
 const promptDelete = ({
-  cli,
   crud,
   name,
 }: {|
-  cli: InteractiveCLI,
   crud: CRUDResource<*, *>,
   name: string,
-|}) =>
-  cli.vorpal.activeCommand.prompt({
```
dicarlo2: Actually, let's add `prompt` to the `InteractiveCLI` object directly, in case we need to change it again in the future. Add a method to `InteractiveCLI` that just delegates to `inquirer.prompt`.

packages/neo-one-cli/src/interactive/createCRUD.js
```diff
@@ -29,6 +29,8 @@ import { timer } from 'rxjs/observable/timer';
 
 import type InteractiveCLI from '../InteractiveCLI';
 
+const inquirer = require('inquirer');
```
dicarlo2: Use `import`, never `require`.

Issue #45, opened 2018-01-02
Migrate old @neo-one/client tests


dicarlo2:
### Description

I reworked the Client apis but kept the old files around under [migrate](/migrate) since there was good test coverage and I think we can reuse a lot of the tests. Let's try to move as much of the testing to the new Client apis and then remove the migrate folder.

PR #64, opened 2018-02-07
Added JSONRPCClient testing

afragapane:
### Description of the Change
Title.

### Applicable Issues
#45


packages/neo-one-client-core/src/__data__/transactions.js
```diff
@@ -0,0 +1,26 @@
+/* @flow */
+import {
+  ASSET_TYPE,
+  RegisterTransaction,
+  ScriptBuilder,
+  common,
+  crypto,
+} from '../';
```
dicarlo2: Move this to `neo-one-client` and change this line to `} from '@neo-one-client-core'`

packages/neo-one-client/src/__tests__/provider/neoone/JSONRPCClient.test.js
```diff
@@ -0,0 +1,161 @@
+/* @flow */
+import BigNumber from 'bignumber.js';
+
+import { keys, transactions } from '@neo-one/client-core/src/__data__';
+
+import JSONRPCClient from '../../../provider/neoone/JSONRPCClient';
+import { RelayTransactionError } from '../../../errors';
+
+jest.mock('isomorphic-fetch');
```
dicarlo2: I think this can be removed, we're just forwarding calls to the underlying provider and no `fetch` is happening.

packages/neo-one-client/src/__tests__/provider/neoone/JSONRPCClient.test.js
```diff
@@ -0,0 +1,161 @@
+/* @flow */
+import BigNumber from 'bignumber.js';
+
+import { keys, transactions } from '@neo-one/client-core/src/__data__';
```
dicarlo2: Yea, typically not good to assume the structure of an external module. Instead, let's copy the `keys` file to the `__data__` directory inside `neo-one-client`. It's definitely duplication, but in this case it's a bit cleaner than trying to import like this.

packages/neo-one-client/src/__tests__/provider/neoone/JSONRPCClient.test.js
```diff
@@ -0,0 +1,161 @@
...
+      method: 'testInvocation',
+      args: [Buffer.alloc(1, 0)],
+    },
+    {
+      method: 'getTransactionReceipt',
+      args: [keys[0].scriptHash],
+    },
+    {
+      method: 'getTransactionReceipt',
+      args: [keys[0].scriptHash],
```
dicarlo2: Should also test the case where options is non-null (so `watchTimeout` is defined in the snapshot, see line 183 in the snapshot file)

packages/neo-one-client/src/__tests__/provider/neoone/JSONRPCClient.test.js
```diff
@@ -0,0 +1,161 @@
...
+      method: 'getAsset',
+      args: [transactions.register.hash],
+    },
+    {
+      method: 'getBlock',
+      args: [transactions.register.hash],
+    },
+    {
+      method: 'getBlock',
+      args: [0],
```
dicarlo2: Same here as `getTransactionReceipt`, add a case where `options` is non-null so `watchTimeout` is defined in the snapshot.

packages/neo-one-client/src/__data__/index.js
```diff
@@ -0,0 +1,4 @@
+/* @flow */
+// eslint-disable-next-line
```
dicarlo2: remove

packages/neo-one-client/src/__tests__/provider/neoone/NEOONEDataProvider.test.js
```diff
@@ -0,0 +1,810 @@
...
+import { AsyncIterableX } from 'ix/asynciterable/asynciterablex';
+import BigNumber from 'bignumber.js';
+
+import { toArray } from 'ix/asynciterable/toarray';
+
+import NEOONEDataProvider from '../../../provider/neoone/NEOONEDataProvider';
+
+import { keys, transactions } from '../../../__data__';
+
+function genExpectedInvocationResult(options: Object) {
```
dicarlo2: In some languages, e.g. php/hack, prefixing a method with `gen` indicates it's an async method. We don't follow that convention (and generally it's not done in JavaScript), but it'd be best to avoid using that prefix to avoid confusion. Instead, perhaps prefix with `create`. Same with all the other methods.

packages/neo-one-client/src/__tests__/provider/neoone/NEOONEDataProvider.test.js
```diff
@@ -0,0 +1,810 @@
...
+        expected: genExpectedTransaction('PublishTransaction', {
+          contract: genExpectedContract({}),
+        }),
+        transactionJSON: genTransactionJSON('PublishTransaction', {
+          contract: genContractJSON({}),
+        }),
+      },
+      {
+        expected: genExpectedTransaction('RegisterTransaction', {
+          asset: (({ type, name, amount, precision, owner, admin }) => ({
```
dicarlo2: Extract this logic out into a common function, like `createRegisterExpectedAsset`.

dicarlo2: Also, as a small nit, it took me a second to figure out what was going on here - creating an anonymous function just to extract some of the properties is not a very common pattern.

packages/neo-one-client/src/__tests__/provider/neoone/NEOONEDataProvider.test.js
```diff
@@ -0,0 +1,810 @@
...
+    );
+
+    const result = provider.getTransactionReceipt(transactions.register.hash);
+    expect(result).resolves.toEqual(expected);
+  });
+
+  test('getInvocationData', async () => {
+    const expected = genExpectedInvocationData({});
+
+    // JSONHelper.readFixed8 = jest.fn(() => 0);
```
dicarlo2: Remove

packages/neo-one-client/src/__tests__/provider/neoone/NEOONEDataProvider.test.js
```diff
@@ -0,0 +1,810 @@
...
+    // $FlowFixMe
+    provider._client.getNetworkSettings = jest.fn(() =>
+      Promise.resolve({ issueGASFee: '0' }),
+    );
+
+    const result = provider.getNetworkSettings();
+    expect(result).resolves.toEqual(expected);
+  });
+
+  test('_ConvertConfirmedTransaction throws null data error', async () => {
```
dicarlo2: nit: `_convertConfirmedTransaction`

PR #69, opened 2018-02-20
Add LocalUserAccountProvider testing

afragapane:
### Description of the Change
Title.

### Applicable Issues
#45

### DIFF
--- a/packages/neo-one-client/src/__tests__/user/LocalUserAccountProvider.test.js
+++ b/packages/neo-one-client/src/__tests__/user/LocalUserAccountProvider.test.js
@@ -1,0 +1,1099 @@
+/* @flow */
+import BigNumber from 'bignumber.js';
+
+import { common, utils } from '@neo-one/client-core';
+
+import LocalUserAccountProvider from '../../user/LocalUserAccountProvider';
+import * as clientUtils from '../../utils';
+import {
+  InsufficientFundsError,
+  InvalidTransactionError,
+  InvokeError,
+  NoAccountError,
+  NothingToClaimError,
+  NothingToTransferError,
+} from '../../errors';
+
+describe('LocalUserAccountProvider', () => {
+  const id1 = {
+    network: 'net1',
+    address: 'ALq7AWrhAueN6mJNqk6FHJjnsEoPRytLdW',
+  };
+  const account1 = {
+    type: 'test',
+    id: id1,
+    name: 'name1',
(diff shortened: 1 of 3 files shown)


packages/neo-one-client/src/__tests__/user/LocalUserAccountProvider.test.js
```diff
@@ -0,0 +1,1118 @@
...
+      wallets.push(wallet);
+      return Promise.resolve();
+    },
+    deleteWallet: wallet => {
+      wallets.splice(wallets.indexOf(wallet), 1);
+      return Promise.resolve();
+    },
+  };
+
+  let keystore = new LocalKeyStore({ store });
```
dicarlo2: Looks all the methods called on these are being mocked out, any particular reason to instantiate a real version here? You could just use an empty object.

packages/neo-one-client/src/__tests__/user/LocalUserAccountProvider.test.js
```diff
@@ -0,0 +1,1118 @@
...
+    const expected = [account1];
+    // $FlowFixMe
+    keystore.getAccounts = jest.fn(() => expected);
+
+    const result = localUserAccountProvider.getAccounts();
+    expect(result).toEqual(expected);
+  });
+
+  test('getNetworks', () => {
+    const expected = '10';
```
dicarlo2: Make this return the correct type. If we change the implementation in the future in an interface backwards compatible way, this test might still fail. E.g. if we try to filter the return of `getAccounts`. Looks like this might be the only spot that needs fixing, the rest look fine, or you're using dummy values where it would be painful and unnecessary to return the correct type.

packages/neo-one-client/src/__tests__/user/LocalUserAccountProvider.test.js
```diff
@@ -0,0 +1,1118 @@
...
+    const result = localUserAccountProvider.getNetworks();
+    expect(result).toEqual(expected);
+  });
+
+  test('selectAccount', async () => {
+    // $FlowFixMe
+    keystore.selectAccount = jest.fn(() => Promise.resolve());
+
+    await localUserAccountProvider.selectAccount(id1);
+    expect(keystore.selectAccount).toHaveBeenCalledTimes(1);
```
dicarlo2: I like to use snapshot tests for these, something like `expect(keystore.selectAccount.mock.call).toMatchSnapshot()`

PR #79, opened 2018-02-28
More VM tests

afragapane:
### Description of the Change
The rest of the non-error test cases for the VM. 
1 bug fix:
Added method to StackItemIterator: .asIterator()


packages/neo-one-node-vm/src/__tests__/opcodes.test.js
```diff
@@ -870,16 +914,30 @@ const OPCODES = ([
...
       args: [keys[0].publicKey, Buffer.alloc(64, 10)],
       result: [new BooleanStackItem(false)],
       gas: FEES.ONE_HUNDRED,
     },
-    // {
-    //   op: 'CHECKMULTISIG',
-    // },
+    {
+      op: 'CHECKMULTISIG',
+      args: [[keys[0].publicKey, keys[1].publicKey], [signature0, signature1]],
```
dicarlo2: Can you test this where we have say 3 public keys and 2 signatures, where the first and third public key match the signatures?

packages/neo-one-node-vm/src/__tests__/opcodes.test.js
```diff
@@ -870,16 +914,30 @@ const OPCODES = ([
...
+      result: [new BooleanStackItem(true)],
+      gas: FEES.ONE_HUNDRED.mul(new BN(2)),
+    },
+    {
+      op: 'CHECKMULTISIG',
+      args: [[keys[0].publicKey, keys[1].publicKey], [Buffer.alloc(64, 10)]],
+      result: [new BooleanStackItem(false)],
+      gas: FEES.ONE_HUNDRED.mul(new BN(2)),
+    },
```
dicarlo2: Also, can you test the case where the inputs are not arrays?

packages/neo-one-node-vm/src/__tests__/opcodes.test.js
```diff
@@ -1135,12 +1193,20 @@ describe('opcodes', () => {
           }),
         ],
       });
+      transaction._message = jest.fn(() => Buffer.alloc(32, 10));
```
dicarlo2: Same here, move to a `mock` key.

packages/neo-one-node-vm/src/__tests__/syscalls.test.js
```diff
@@ -98,19 +98,21 @@ const asset = {
...
+const iterator = AsyncIterableX.of();
+const nextItem = {
+  value: new StorageItem({
+    hash: scriptAttributeHash,
+    key: Buffer.from('key', 'utf-8'),
+    value: Buffer.from('val', 'utf-8'),
+  }),
+};
+// $FlowFixMe
+iterator.next = jest.fn(() => Promise.resolve(nextItem));
```
dicarlo2: Just use `AsyncIterableX.of(nextItem)`?

packages/neo-one-node-vm/src/__tests__/syscalls.test.js
```diff
@@ -98,19 +98,21 @@ const asset = {
...
+const stackIterator = new StackItemIterator((iterator: $FlowFixMe));
+async function initializeIterator() {
+  await stackIterator.next();
+}
+initializeIterator();
```
dicarlo2: This causes a race condition with the tests since we're not waiting for it to complete. It's possible the test runs before `initializeIterator` is actually called. Add a key to `SysCall` called like `pretest` that is an optional async function to execute before the test executes.

Issue #80, opened 2018-02-28
Add `bootstrap` command to `neo-one-server-plugin-network`


dicarlo2:
This command should bootstrap a private network node with a bunch of data. To start, let's bootstrap with
 - ContractTransactions (`client.transfer`)
 - Register a couple of assets (`client.registerAsset`)

Basically you'll want to add the `master` wallet that's automatically created with the private network to the `client` along with creating some other wallets to transfer and register assets with. That should be relatively straightforward for the first PR. I actually started on this, here's a gist of a WIP: https://gist.github.com/dicarlo2/4b6f3234c95443a570786828a38508d2. You'll just need to add it as an interactive command to the network plugin and then fill in the file.

Next, we'll want to add more transactions:
 - ClaimTransactions (`client.claim`)
 - Issue coins for the assets created above (`client.issue`)

However, this requires the previous transactions to be confirmed in a block. And waiting for a block to be created every time we need one will make this command take a long time. So, before adding the above transactions, let's add a new client type to `neo-one-client`: `DeveloperClient`. This client should work similar to `ReadClient` in that it talks to a single network (unlike `Client` which can talk to many networks) and takes a new `DeveloperProvider` type. It should have the following methods (which basically just delegate to `DeveloperProvider`:
 - `startConsensus()`: triggers consensus on the node immediately
 - `updateOptions(options: Options)`: updates various options for the node. Initially, it will just have one option `secondsPerBlock` which should update the `blockchain.settings.secondsPerBlock`. Since we're all about immutability, you'll probably want to change `settings` in `Blockchain` to be a `BehaviorSubject` and add a method `get settings()` which just takes the current value so it's backwards compatible. That way you can update it.
 - `fastForward(seconds: number)`: fast forward the block time by `seconds` seconds. Take a look at `runConsensus` to see where you need to feed this value in order to force the next block to use a time in the future.

Each method should probably be it's own PR since they're relatively independent and will require a decent amount of work to thread the values from `neo-one-client`, through `neo-one-node-rpc` to their final destination. For the `bootstrap` command we'll likely only use `startConsensus`, but the other 2 will come in handy later. Remember to add tests so we keep `neo-one-client` at near 100% coverage!

Once that's done, we can easily add the other transactions mentioned above in another PR. Spread the transactions out over a few blocks using `startConsensus` so that we have a good number of blocks and transactions.

Finally, let's add a few NEP-5 contracts:
[...]

PR #82, opened 2018-03-02
Add bootstrap command to wallet plugin

afragapane:
### Requirements

### Description of the Change
Basic bootstrap command added to wallet plugin.  At this time, the command will bootstrap a network with master wallet and a configurable number of other wallets (default 10) with a random distribution of funds.  

### Applicable Issues
#80


packages/neo-one-server-client/src/Client.js
```diff
@@ -151,6 +173,40 @@ export default class Client {
...
+      options,
+      cancel$,
+    }).toPromise();
+
+    const error = getTasksError(response.tasks);
+    if (error != null) {
+      throw new Error(error);
+    }
+
+    return this.getResource({
```
dicarlo2: Should probably `await` this and then enforce it's not null so that `createResource` can return a non-null value.

packages/neo-one-server-plugin-wallet/src/__e2e__/bootstrap.test.js
```diff
@@ -0,0 +1,78 @@
...
+import { addressToScriptHash } from '@neo-one/client';
+
+describe('bootstrap', () => {
+  test.only('bootstrap on active network - default', async () => {
+    const numWallets = 10;
+
+    await one.execute('create network boottest');
+    await new Promise(resolve => setTimeout(() => resolve(), 10000));
+
+    await one.execute('bootstrap');
```
dicarlo2: Add `--network boottest`

packages/neo-one-server-plugin-wallet/src/__e2e__/bootstrap.test.js
```diff
@@ -0,0 +1,78 @@
...
+    await new Promise(resolve => setTimeout(() => resolve(), 10000));
+
+    const output = await one.execute('get wallet --json');
```
dicarlo2: Same here

dicarlo2: And everywhere else in this file. Remember, this is running as just a normal command, not from the interactive cli, so no network is activated.

packages/neo-one-server-plugin-wallet/src/bootstrap.js
```diff
@@ -0,0 +1,287 @@
...
+} from '@neo-one/client';
+import { common } from '@neo-one/client-core';
+import {
+  constants as networkConstants,
+  type Network,
+} from '@neo-one/server-plugin-network';
+
+import _ from 'lodash';
+import { of as _of } from 'rxjs/observable/of';
+import BigNumber from 'bignumber.js';
```
dicarlo2: nit: imports should be organized like
- Capital letter module imports
- space
- Lower case letter module imports
- space
- Capital letter relative imports
- space
- Lower case letter relative imports

packages/neo-one-server-plugin-wallet/src/bootstrap.js
```diff
@@ -0,0 +1,287 @@
...
+    resourceType: constants.WALLET_RESOURCE_TYPE,
+    name: constants.makeWallet({ name: walletName, network: networkName }),
+    options: {
+      network: networkName,
+    },
+    cancel$: _of(),
+  });
+
+  const wallet = ((walletResource: $FlowFixMe): Wallet);
+  if (wallet == null) {
```
dicarlo2: Move this check to inside `createResource`

packages/neo-one-server-plugin-wallet/src/bootstrap.js
```diff
@@ -0,0 +1,287 @@
...
+    privateKey: wifToPrivateKey(wallet.wif),
+  });
+}
+
+function getNumWallets(options): number {
+  const { wallets } = options;
+  if (wallets != null && typeof wallets === 'number') {
+    return wallets;
+  }
+  return DEFAULT_NUM_WALLETS;
```
dicarlo2: Should throw an error if `wallets` is provided and it's not a number.

packages/neo-one-server-plugin-wallet/src/bootstrap.js
```diff
@@ -0,0 +1,287 @@
...
+          createTransfers(walletName, network.name, cli, plugin),
+        ),
+      );
+      transfersBatch1 = _.flatten(transfersBatch1);
+
+      const transactionBatch1 = await client.transfer(transfersBatch1, {
+        from: masterWallet.accountID,
+      });
+      await transactionBatch1.confirmed();
+      await new Promise(resolve => setTimeout(() => resolve(), 2000));
```
dicarlo2: Remove, see my comment above.

packages/neo-one-server-plugin-wallet/src/bootstrap.js
```diff
@@ -0,0 +1,287 @@
...
+        memory: new LocalUserAccountProvider({
+          keystore,
+          provider: new NEOONEProvider({
+            options: [
+              { network: network.name, rpcURL: network.nodes[0].rpcAddress },
+            ],
+          }),
+        }),
+      });
+      const [walletsBatch1, walletsBatch2] = _.chunk(
```
dicarlo2: nit: avoid appending with numbers, use full names like `firstBatch` and `secondBatch`, it's much easier to visually and mentally distinguish. Similar to not using abbreviations for variable names.

packages/neo-one-server-plugin-wallet/src/bootstrap.js
```diff
@@ -0,0 +1,287 @@
...
+
+  let neo;
+  let gas;
+  if (from == null) {
+    const seed = randomInt(100);
+    neo = randomIntDist(seed);
+    gas = randomIntDist(seed);
+  } else {
+    const transferPercent = randomInt(75) / 100;
+    neo = Math.ceil(transferPercent * Number(from.neoBalance));
```
dicarlo2: The problem is that `from.neoBalance` is updated via polling, whereas `confirmed` is updated instantaneously when the block is confirmed. So there's a race condition where the `client` knows the block has been confirmed and the neo-one server actually updates the balance. Instead, use `client.read(<networkName>).getAccount` to get the current balance directly from the blockchain rather than the neo-one server.

packages/neo-one-server-plugin-wallet/src/__e2e__/bootstrap.test.js
```diff
@@ -0,0 +1,52 @@
...
+
+describe('bootstrap', () => {
+  test('bootstrap - default', async () => {
+    const numWallets = 10;
+
+    await one.execute('create network boottest');
+    await new Promise(resolve => setTimeout(() => resolve(), 10000));
+
+    await one.execute('bootstrap --network boottest');
+    await new Promise(resolve => setTimeout(() => resolve(), 10000));
```
dicarlo2: I think you can remove this `setTimeout` since the command should only complete once it's fully done. Even the `create network` one might not be necessary anymore because we wait for the node to start up before finishing the command. In the network e2e IIRC I added it before we were waiting for the node to start up.

packages/neo-one-server-plugin-wallet/src/__e2e__/bootstrap.test.js
```diff
@@ -0,0 +1,52 @@
...
+
+    const output = await one.execute('get wallet --network boottest --json');
+    let wallets = one.parseJSON(output);
+
+    expect(wallets.length).toEqual(numWallets + 2);
```
dicarlo2: Why `numWallets + 2`?  Add a comment.

packages/neo-one-server-plugin-wallet/src/__e2e__/bootstrap.test.js
```diff
@@ -0,0 +1,52 @@
...
+    wallets = wallets.slice(1);
+
+    for (const wallet of wallets) {
+      expect(wallet[0]).toEqual('boottest');
+      expect(addressToScriptHash(wallet[2])).toBeDefined();
+      expect(wallet[3]).toEqual('Yes');
+      expect(Number(wallet[4])).toBeGreaterThan(0);
```
dicarlo2: Can you add a comment to each of these to remind us what we're testing here? E.g.
```
// wallet.neoBalance
expect(Number(wallet[4])).toBeGreaterThan(0);
```

packages/neo-one-server-plugin-wallet/src/__e2e__/bootstrap.test.js
```diff
@@ -0,0 +1,52 @@
...
+      expect(Number(wallet[5])).toBeGreaterThan(0);
+    }
+  });
+
+  test('bootstrap - 20 wallets', async () => {
```
dicarlo2: Could probably extract out the common logic here, at least the final `expect` calls.

packages/neo-one-server-plugin-wallet/src/bootstrap.js
```diff
@@ -0,0 +1,325 @@
...
+    name: wallet.name,
+    privateKey: wifToPrivateKey(wallet.wif),
+  });
+}
+
+function getNumWallets(options): number {
+  const { wallets } = options;
+  if (wallets != null && typeof wallets === 'number') {
+    return wallets;
+  } else if (wallets != null && typeof wallets !== 'number') {
```
dicarlo2: DeepScan shows an error here - take a look at the PR checks.

packages/neo-one-server-plugin-wallet/src/bootstrap.js
```diff
@@ -0,0 +1,325 @@
...
+  networkName: string,
+  cli: InteractiveCLI,
+  plugin: WalletPlugin,
+  from?: Account,
+|}): Promise<Array<Transfer>> {
+  const wallet = await getWallet({ walletName, networkName, cli, plugin });
+
+  let neo;
+  let gas;
+  if (from != null) {
```
dicarlo2: Super nitpicky, but I prefer having the positive condition always go first, i.e. `if (from == null) {`.

packages/neo-one-server-plugin-wallet/src/bootstrap.js
```diff
@@ -0,0 +1,325 @@
...
+  }
+
+  await keystore.addAccount({
+    network: networkName,
+    name: wallet.name,
+    privateKey: wifToPrivateKey(wallet.wif),
+  });
+}
+
+function getNumWallets(options): number {
```
dicarlo2: Add a type here (I think it should just be `options: Object`).

dicarlo2:
Also looks like there's a failure in the e2e tests, but it seems like it passed in appveyor. Looks like it has something to do with transferring, make sure that the e2e tests consistently pass, otherwise it's going to cause a lot of churn/pain with spurious failures.

PR #110, opened 2018-03-23
Add BrowserMonitor

afragapane:
### Description of the Change
Filled out BrowserMonitor to allow for monitoring of the the browser.  This was done by adding:  
BrowserLogger: Collects raw logs, reports & resets on command
BrowserMetricsFactory: Collects & keeps track of created metrics.  Reports & resets on command.
Reporter: Collects logs & metrics from BrowserLogger & BrowserMetricsFactory at a regular set time interval.  Sends these to a pre-configured endpoint.  If this fails, will keep a backlog of up to 1000 records for each log or metric type.
MonitorBase: Ability to receive logs & metrics and distribute them to the appropriate location for logging in the NodeMonitor.

### Test Plan
Unit tests added for all of the above.


packages/neo-one-monitor/src/BrowserMonitor.js
```diff
@@ -8,14 +21,26 @@ import MonitorBase, {
   type Tracer,
 } from './MonitorBase';
 
+const MAX_BACKLOG = 1000;
+
 class BaseMetric {
-  // eslint-disable-next-line
-  constructor(options: MetricConstruct) {}
+  metric: MetricConstruct;
+  values: Array<Object>;
```
dicarlo2: Type this properly

packages/neo-one-monitor/src/BrowserMonitor.js
```diff
@@ -73,7 +261,7 @@ export default class BrowserMonitor extends MonitorBase {
     metricsLogLevel,
     spanLogLevel,
   }: BrowserMonitorCreate): BrowserMonitor {
-    const perf = performance;
+    const perf = perfHooks.performance;
```
dicarlo2: Revert, this should be how it was. `perf_hooks` is a node only module.

packages/neo-one-monitor/src/BrowserMonitor.js
```diff
@@ -30,30 +55,193 @@ class BrowserGauge extends BaseMetric implements Gauge {
...
+
+    return logs;
+  }
+
+  close(callback: () => void): void {
+    callback();
+  }
+}
+
+export class Reporter {
```
dicarlo2: Move this into it's own file

packages/neo-one-monitor/src/BrowserMonitor.js
```diff
@@ -30,30 +55,193 @@ class BrowserGauge extends BaseMetric implements Gauge {
...
+}
+
+type MetricCollection = {|
+  counters: Array<BaseMetric>,
+  gauges: Array<BaseMetric>,
+  histograms: Array<BaseMetric>,
+  summaries: Array<BaseMetric>,
+|};
+
+class BrowserMetricsFactory implements MetricsFactoryCollect {
```
dicarlo2: Move this into it's own file and leave the old factory and metrics how they were. I.e. this file should basically have no changes.

dicarlo2: The reason being is we shouldn't force all consumers to use the metrics that are collected and reported - instead they can use the default no-op implementation.

packages/neo-one-monitor/src/MonitorBase.js
```diff
@@ -721,6 +729,38 @@ export default class MonitorBase implements Span {
     }
   }
 
+  receive(report: Report): void {
```
dicarlo2: Maybe call this `report` since I suppose it could be used in either the browser or node for bulk metric addition.

packages/neo-one-monitor/src/BrowserMonitor.js
```diff
@@ -30,30 +55,193 @@ class BrowserGauge extends BaseMetric implements Gauge {
...
+    this._endpoint = endpoint;
+    this._backLogs = [];
+    this._backMetrics = {
+      counters: [],
+      gauges: [],
+      histograms: [],
+      summaries: [],
+    };
+
+    const subscription = interval(this._timer).subscribe(async () => {
```
dicarlo2: `this._subscription = ` and `this._subscription.unsubscribe` in `close` below.

packages/neo-one-monitor/src/BrowserMonitor.js
```diff
@@ -30,30 +55,193 @@ class BrowserGauge extends BaseMetric implements Gauge {
...
+  }: {|
+    logger: BrowserLogger,
+    monitor: BrowserMonitor,
+    timer: number,
+    endpoint: string,
+  |}) {
+    this._logger = logger;
+    this._monitor = monitor;
+    this._timer = timer;
+    this._endpoint = endpoint;
```
dicarlo2: Only this one needs to be a instance property, all of the other can stay local to the constructor.

packages/neo-one-monitor/src/BrowserMonitor.js
dicarlo2: use `pipe(concatMap(async () => {})).subscribe()` to ensure that the timers are processed serially. As it is now, it's possible that two invocations of your subscribe function happen simultaneously. In general, you almost never want to use `subscribe` directly.

dicarlo2: In fact, it might be a good exercise to try to translate this to use `.pipe(mergeScan(..., [], 1))` and eliminate the local state changes - only use the aggregated argument for `mergeScan`.

packages/neo-one-monitor/src/BrowserMonitor.js
```diff
@@ -30,30 +55,193 @@ class BrowserGauge extends BaseMetric implements Gauge {
...
+    this._metrics = {
+      counters: [],
+      gauges: [],
+      histograms: [],
+      summaries: [],
+    };
+  }
+
+  collect(): MetricCollection {
+    const metrics = JSON.parse(JSON.stringify(this._metrics));
```
dicarlo2: This is a janky pattern. Makes a lot of assumptions about how `BaseMetric` is implemented. Instead, you should add an explicit `serializeJSON` method.

packages/neo-one-monitor/src/BrowserMonitor.js
```diff
@@ -30,30 +55,193 @@ class BrowserGauge extends BaseMetric implements Gauge {
...
+  gauges: Array<BaseMetric>,
+  histograms: Array<BaseMetric>,
+  summaries: Array<BaseMetric>,
+|};
+
+class BrowserMetricsFactory implements MetricsFactoryCollect {
+  _metrics: MetricCollection;
+
+  constructor() {
+    this._metrics = {
```
dicarlo2: Make this at the module scope like `NodeMonitor` so that we enforce there's only ever one instance of each metric per application.

packages/neo-one-monitor/src/BrowserMonitor.js
```diff
@@ -30,30 +55,193 @@ class BrowserGauge extends BaseMetric implements Gauge {
...
+
+    const subscription = interval(this._timer).subscribe(async () => {
+      let logs = this._logger.collect();
+      const metrics = this._monitor.collect();
+
+      if (this._backLogs.length > 0) {
+        logs = logs.concat(this._backLogs);
+        this._backLogs = [];
+
+        for (const metricType of Object.keys(metrics)) {
```
dicarlo2: This is a potentially expensive loop (`O(n^2)`) if there are a lot of metrics. Instead, you should refactor this so that you can just do a lookup to see if `_backMetrics` has the `name` you're looking for. I.e. `_backMetrics` should be something like an object from type => name => metric

packages/neo-one-monitor/src/BrowserMonitor.js
```diff
@@ -30,30 +55,193 @@ class BrowserGauge extends BaseMetric implements Gauge {
...
+        }
+        this._backMetrics = {
+          counters: [],
+          gauges: [],
+          histograms: [],
+          summaries: [],
+        };
+      }
+
+      const response = await this._report({ logs, metrics });
```
dicarlo2: Need to catch errors here too. Also would be good to increment a counter for requests with a label for whether the request failed or not.

packages/neo-one-monitor/src/BrowserMonitor.js
```diff
@@ -90,4 +278,9 @@ export default class BrowserMonitor extends MonitorBase {
       spanLogLevel,
     });
   }
+
+  collect(): MetricCollection {
```
dicarlo2: Remove, this should only be on the metrics factory itself which gets passed into `create` by the caller.

packages/neo-one-monitor/src/MonitorBase.js
```diff
@@ -721,6 +729,38 @@ export default class MonitorBase implements Span {
...
+    report.logs.map(log => this._logger.log(log));
+
+    for (const counterMetric of report.metrics.counters) {
+      const counter = this.getCounter({
+        name: counterMetric.metric.name,
+        help: counterMetric.metric.help,
+        labelNames: counterMetric.metric.labelNames,
+        metricLabels: counterMetric.metric.metricLabels,
+        receivedFromBrowser: true,
```
dicarlo2: What does this do?

packages/neo-one-monitor/src/BrowserMonitor.js
```diff
@@ -30,30 +55,193 @@ class BrowserGauge extends BaseMetric implements Gauge {
...
+  }
+
+  close(callback: () => void): void {
+    callback();
+  }
+}
+
+export class Reporter {
+  _logger: BrowserLogger;
+  _monitor: BrowserMonitor;
```
dicarlo2: This should be the metrics factory instead.

packages/neo-one-monitor/src/BrowserMetricsFactory.js
```diff
@@ -0,0 +1,123 @@
...
+    metrics.histograms.push(histogram);
+
+    return histogram;
+  }
+
+  createSummary(options: MetricConstruct): Summary {
+    return new BrowserSummary(options);
+  }
+
+  _serializeJSON(inputJSON: Object): Object {
```
dicarlo2: This isn't quite what I meant by `_serializeJSON`. You should have an explicit type in and an explicit type out - not trying to implement some kind of generic jsonify function. Do something like
```
_serializeJSON(metrics: Metrics): MetricCollection
```

This also is totally broken - you're returning `MetricCollection` from `collect` when it's not matching the type at all.

packages/neo-one-monitor/src/BrowserMonitor.js
```diff
@@ -18,7 +19,7 @@ class BrowserCounter extends BaseMetric implements Counter {
   inc(countOrLabels?: number | MetricLabels, count?: number): void {}
 }
 
-class BrowserGauge extends BaseMetric implements Gauge {
+export class BrowserGauge extends BaseMetric implements Gauge {
```
dicarlo2: Is this being used somewhere?

packages/neo-one-monitor/src/BrowserMonitor.js
```diff
@@ -1,5 +1,6 @@
 /* @flow */
 import type { Counter, Gauge, Histogram, LogLevel, Summary } from './types';
+
```
dicarlo2: Revert

packages/neo-one-monitor/src/BrowserMonitor.js
```diff
@@ -69,11 +71,12 @@ export default class BrowserMonitor extends MonitorBase {
   static create({
     service,
     logger,
+    metricsFactory,
     tracer,
     metricsLogLevel,
     spanLogLevel,
   }: BrowserMonitorCreate): BrowserMonitor {
-    const perf = performance;
+    const perf = global.performance;
```
dicarlo2: Rebase

packages/neo-one-monitor/src/BrowserLogger.js
```diff
@@ -0,0 +1,31 @@
...
+
+  constructor() {
+    this._logs = [];
+  }
+
+  log(options: LoggerLogOptions): void {
+    this._logs.push(options);
+  }
+
+  collect(): Array<LoggerLogOptions> {
```
dicarlo2: This doesn't work for the `error` key.

packages/neo-one-monitor/src/MonitorBase.js
```diff
@@ -777,6 +785,38 @@ export default class MonitorBase implements Span {
...
+      });
+
+      counterMetric.values.forEach(value => {
+        counter.inc(value.countOrLabels, value.count);
```
dicarlo2: This isn't going to work anymore with the latest changes (might need to rebase). `getCounter` returns a wrapped metric that adds the current monitor's labels.

packages/neo-one-monitor/src/MonitorBase.js
```diff
@@ -777,6 +785,38 @@ export default class MonitorBase implements Span {
...
+      const histogram = this.getHistogram({
+        name: histMetric.metric.name,
+        help: histMetric.metric.help,
+        labelNames: histMetric.metric.labelNames,
+        metricLabels: histMetric.metric.metricLabels,
+        receivedFromBrowser: true,
+      });
+
+      histMetric.values.forEach(value => {
+        histogram.observe(value.countOrLabels, value.count);
```
dicarlo2: Same here

packages/neo-one-monitor/src/MonitorBase.js
dicarlo2: Probably want to disable returning the wrapper when `receivedFromBrowser` is true.

packages/neo-one-monitor/src/MonitorBase.js
dicarlo2: I'd probably call this something like `directReport`.

packages/neo-one-monitor/src/Reporter.js
```diff
@@ -0,0 +1,139 @@
...
+  }
+
+  async _report(report: Report): Promise<Response> {
+    return fetch(this._endpoint, {
+      method: 'POST',
+      body: JSON.stringify(report),
+    });
+  }
+
+  _checkBackReport(report: Report, backReport: BackReport): Report {
```
dicarlo2: `_addBackReport`?

packages/neo-one-monitor/src/Reporter.js
```diff
@@ -0,0 +1,139 @@
...
+            let report = { logs, metrics };
+
+            if (backReport != null) {
+              report = this._checkBackReport({ logs, metrics }, backReport);
+            }
+            let response;
+            try {
+              response = await this._report(report);
+            } catch (error) {
+              throw new Error(error);
```
dicarlo2: Why are we catching an error just to rethrow it? Also `error` is an `Error` instance which doesn't make sense to construct another `Error` with.

packages/neo-one-monitor/src/Reporter.js
```diff
@@ -0,0 +1,139 @@
...
+            }
+            if (!response.ok) {
+              requestCounter.inc({ failure: 1 });
```
dicarlo2: use `[KNOWN_LABELS.ERROR]: true` and `[KNOWN_LABELS.ERROR]: false`

packages/neo-one-monitor/src/Reporter.js
```diff
@@ -0,0 +1,139 @@
...
+
+  _checkBackReport(report: Report, backReport: BackReport): Report {
+    const logs = report.logs.concat(backReport.logs);
+    const { metrics } = report;
+
+    for (const metricType of Object.keys(metrics)) {
+      for (const metric of metrics[metricType]) {
+        if (metric != null) {
+          const { name } = metric.metric;
+          if (name in backReport.metrics[metricType]) {
```
dicarlo2: Don't use `in` - `backReport.metrics[metricType][name] != null`

packages/neo-one-monitor/src/Reporter.js
```diff
@@ -0,0 +1,139 @@
...
+            metric.values = metric.values.concat(
```
dicarlo2: This messes up the order of the values, you're `push`ing above (i.e. adding to the end of the array) and then forcing the current reports values to be added to the front of the array.

packages/neo-one-monitor/src/Reporter.js
```diff
@@ -0,0 +1,139 @@
...
+              backReport.metrics[metricType][name].values,
+            );
+          } else {
+            metrics[metricType].push(backReport.metrics[metricType][name]);
```
dicarlo2: Why are we pushing `undefined` values onto the array?

packages/neo-one-monitor/src/Reporter.js
```diff
@@ -0,0 +1,139 @@
...
+      body: JSON.stringify(report),
+    });
+  }
+
+  _checkBackReport(report: Report, backReport: BackReport): Report {
+    const logs = report.logs.concat(backReport.logs);
+    const { metrics } = report;
+
+    for (const metricType of Object.keys(metrics)) {
+      for (const metric of metrics[metricType]) {
```
dicarlo2: This should be using `backReport.metrics` and below you should be checking if `name` exists in `metrics`.

packages/neo-one-monitor/src/Reporter.js
```diff
@@ -0,0 +1,139 @@
...
+
+  async _report(report: Report): Promise<Response> {
+    return fetch(this._endpoint, {
+      method: 'POST',
+      body: JSON.stringify(report),
+    });
+  }
+
+  _checkBackReport(report: Report, backReport: BackReport): Report {
+    const logs = report.logs.concat(backReport.logs);
```
dicarlo2: Same thing here, this messes up the order of the logs - `backReport.logs` should come first.

packages/neo-one-monitor/src/Reporter.js
```diff
@@ -0,0 +1,139 @@
...
+    }
+
+    return {
+      logs,
+      metrics,
+    };
+  }
+
+  _constructBackReport(report: Report): BackReport {
+    const backLogs = report.logs.slice(0, MAX_BACKLOG);
```
dicarlo2: Based on the comments above, this should be `report.logs.slice(-MAX_BACKLOG)`

packages/neo-one-monitor/src/Reporter.js
dicarlo2: Don't mutate `metric.values`.

packages/neo-one-monitor/src/types.js
```diff
@@ -170,6 +181,11 @@ export type CaptureSpanLogOptions = {|
   references?: Array<Reference | void>,
 |};
 
+export type Report = {|
+  logs: Array<LoggerLogOptions>,
+  metrics: Object,
```
dicarlo2: Don't use `Object`

packages/neo-one-monitor/src/BrowserLogger.js
```diff
@@ -0,0 +1,53 @@
...
+    const log = {
+      name: options.name,
+      level: options.level,
+      message: options.message,
+      labels: options.labels,
+      data: options.data,
+      error: undefined,
+    };
+    if (options.error) {
+      log.error = {
```
dicarlo2: Don't mutate, instead do something like
```
error: options.error == null ? undefined : {
  ...
}
```

packages/neo-one-monitor/src/BrowserLogger.js
```diff
@@ -0,0 +1,53 @@
...
+        message: options.error.message,
+        stack: options.error.stack,
```
dicarlo2: You also need to extract the `Error` code if it exists to match `_getErrorKind` - https://github.com/neo-one-suite/neo-one/blob/master/packages/neo-one-monitor/src/MonitorBase.js#L923

packages/neo-one-monitor/src/BrowserMetricsFactory.js
```diff
@@ -0,0 +1,144 @@
...
+
+  _serializeJSON(currentMetrics: MetricCollection): MetricCollection {
+    const copy = {
+      counters: {},
+      histograms: {},
+    };
+
+    for (const metricType of Object.keys(currentMetrics)) {
+      for (const name of Object.keys(currentMetrics[metricType])) {
+        const metricCopy = {
```
dicarlo2: This is another mutation anti-pattern that leads to not well-typed Flow objects. Is there a reason why you're not constructing the object directly? I.e.
```
const metric = {
  metric: {
    name: currentMetrics[metricType][name].metric.name,
    help: currentMetrics[metricType][name].metric.help,
    labelNames: [...currentMetrics[metricType][name].metric.labelNames],
  },
  values: currentMetrics[metricType][name].values.map(...),
}
```
It's much cleaner and clearer to the reader to do it like that.

packages/neo-one-monitor/src/BrowserMetricsFactory.js
```diff
@@ -0,0 +1,144 @@
...
+          values: [],
+        };
+        metricCopy.metric.name = currentMetrics[metricType][name].metric.name;
+        metricCopy.metric.help = currentMetrics[metricType][name].metric.help;
+        metricCopy.metric.labelNames = [];
+        currentMetrics[metricType][name].metric.labelNames.map(label =>
+          metricCopy.metric.labelNames.push(label),
+        );
+        metricCopy.values = [];
+        currentMetrics[metricType][name].values.forEach(value => {
```
dicarlo2: This is a perfect example of what I mean by data transformations vs mutations. As a reader, I now have to look through all of this code line by line to figure out what the overall goal is here of looping over the values. If instead, you told me up front what the goal is by using a transformation function, i.e. `map`, I immediately know we're transforming the current metrics' values. Instead, I have to read through this and more manual iteration to figure that out.

packages/neo-one-monitor/src/BrowserMetricsFactory.js
```diff
@@ -0,0 +1,144 @@
...
+          let countOrLabelsCopy = value.countOrLabels;
+          if (typeof value.countOrLabels === 'object') {
+            countOrLabelsCopy = {};
```
dicarlo2: `countOrLabelsCopy = { ...value.countOrLabels }`

packages/neo-one-monitor/src/BrowserMetricsFactory.js
```diff
@@ -0,0 +1,144 @@
...
+  }
+}
+
+export type MetricCollection = {|
+  counters: { [name: string]: CollectingMetric },
+  histograms: { [name: string]: CollectingMetric },
+|};
+
+class BrowserCounter extends CollectingMetric implements Counter {
+  inc(countOrLabels?: number | MetricLabels, count?: number): void {
```
dicarlo2: I think this should actually always be a `MetricLabels`, can you update the definition of `Counter` (and the other metrics) to reflect that?

packages/neo-one-monitor/src/types.js
```diff
@@ -85,6 +86,15 @@ export type LogSingleOptions = {|
   },
 |};
 
+export type LoggerLogOptions = {|
```
dicarlo2: This should stay in `MonitorBase`, it's not part of the public api of `Monitor`.

packages/neo-one-monitor/src/types.js
```diff
@@ -134,6 +144,8 @@ export type MetricOptions = {|
   name: string,
   help?: string,
   labelNames?: Array<Label>,
+  metricLabels?: Labels,
```
dicarlo2: This should not be part of the public api for `getCounter`, etc. Remove.

packages/neo-one-monitor/src/types.js
```diff
@@ -134,6 +144,8 @@ export type MetricOptions = {|
...
+  directReport?: boolean,
```
dicarlo2: Same here, should not be part of the public api.

packages/neo-one-monitor/src/types.js
dicarlo2: It's ok to import this from `MonitorBase` directly where you need it.

packages/neo-one-monitor/src/MonitorBase.js
```diff
@@ -780,6 +792,73 @@ export default class MonitorBase implements Span {
...
+        if (error.stack != null) {
+          newLog.error.stack = error.stack;
+        }
+      }
+      this._logger.log(newLog);
+    });
+
+    for (const counterName of Object.keys(report.metrics.counters)) {
+      const counterMetric = report.metrics.counters[counterName];
+      const counter = this.getCounter({
```
dicarlo2: Instead of passing in a `directReport` option that changes how `getCounter` functions, just create a separate private method `._getCounter` that both this method and `getCounter` uses that does what you want.

packages/neo-one-monitor/src/MonitorBase.js
```diff
@@ -780,6 +792,73 @@ export default class MonitorBase implements Span {
...
+        name: counterMetric.metric.name,
+        help: counterMetric.metric.help,
+        labelNames: counterMetric.metric.labelNames,
+        directReport: true,
+      });
+
+      counterMetric.values.forEach(value => {
```
dicarlo2: This is a good example of where using `forEach` makes sense - you're doing a side effect (`.inc`) and as such, the iteration should represent that.

packages/neo-one-monitor/src/MonitorBase.js
```diff
@@ -780,6 +792,73 @@ export default class MonitorBase implements Span {
...
+        if (
+          value.countOrLabels != null &&
+          typeof value.countOrLabels === 'object'
+        ) {
+          counter.inc(value.countOrLabels, value.count);
```
dicarlo2: I think this should work without checks, i.e. just do `counter.inc(value.countOrLabels, value.count)` without any checks.

packages/neo-one-monitor/src/MonitorBase.js
```diff
@@ -780,6 +792,73 @@ export default class MonitorBase implements Span {
...
+        directReport: true,
+      });
+
+      histMetric.values.forEach(value => {
+        if (
+          value.countOrLabels != null &&
+          typeof value.countOrLabels === 'object' &&
+          value.count != null
+        ) {
+          histogram.observe(value.countOrLabels, value.count);
```
dicarlo2: Same here, just this line and remove the others.

packages/neo-one-monitor/src/MonitorBase.js
```diff
@@ -780,6 +792,73 @@ export default class MonitorBase implements Span {
...
+          counter.inc(value.countOrLabels, value.count);
+        } else {
+          counter.inc(value.countOrLabels);
+        }
+      });
+    }
+
+    for (const histName of Object.keys(report.metrics.histograms)) {
+      const histMetric = report.metrics.histograms[histName];
+      const histogram = this.getHistogram({
```
dicarlo2: Same comment here.

packages/neo-one-monitor/src/MonitorBase.js
```diff
@@ -934,10 +1013,16 @@ export default class MonitorBase implements Span {
     name,
     help = 'Placeholder',
     labelNames: labelNamesIn,
+    directReport,
   }: MetricOptions): MetricConstruct {
-    const labelNames = this._getMetricLabelNames(labelNamesIn).map(labelName =>
-      convertMetricLabel(labelName),
-    );
+    let labelNames;
+    if (directReport) {
```
dicarlo2: Same comment here, don't change this method to do something entirely different with `directReport`, just make a common private method that they both use.

packages/neo-one-monitor/src/Reporter.js
```diff
@@ -0,0 +1,130 @@
...
+    endpoint,
+  }: {|
+    logger: BrowserLogger,
+    metricsFactory: BrowserMetricsFactory,
+    timer: number,
+    endpoint: string,
+  |}) {
+    this._endpoint = endpoint;
+    const requestCounter = metricsFactory.createCounter({
+      name: 'browserRequestCounter',
```
dicarlo2: `browser_request_counter` (metric names are always snake case)

packages/neo-one-monitor/src/Reporter.js
```diff
@@ -0,0 +1,130 @@
...
+    this._subscription = interval(timer)
+      .pipe(
+        mergeScan(
+          async backReport => {
+            const logs = logger.collect();
+            const metrics = metricsFactory.collect();
+            let report = { logs, metrics };
+
+            if (backReport != null) {
+              report = this._addBackReport({ logs, metrics }, backReport);
```
dicarlo2: `report = this._addBackReport(report, backReport);` (convince yourself why this is possible)

packages/neo-one-monitor/src/Reporter.js
```diff
@@ -0,0 +1,130 @@
...
+    const logs = backReport.logs.concat(report.logs);
+    const { metrics } = report;
+
+    for (const metricType of Object.keys(backReport.metrics)) {
+      for (const name of Object.keys(backReport.metrics[metricType])) {
+        if (metrics[metricType][name] != null) {
+          const newValues = backReport.metrics[metricType][name].values.concat(
+            metrics[metricType][name].values,
+          );
+          metrics[metricType][name].values = newValues;
```
dicarlo2: Same thing here - merge these immutably, i.e. without ever mutating the input metrics. Eliminate the `for` loops and create new objects on each iteration. It's a similar transformation as what you did to make `interval` above immutable.

packages/neo-one-monitor/src/Reporter.js
```diff
@@ -0,0 +1,130 @@
...
+  }
+
+  _constructBackReport(report: Report): Report {
+    const backLogs = report.logs.slice(-MAX_BACKLOG);
+    const backMetrics = {
+      counters: {},
+      histograms: {},
+    };
+
+    for (const metricType of Object.keys(report.metrics)) {
```
dicarlo2: Same here - make this immutable using `.reduce`.

package.json
```diff
@@ -148,5 +148,8 @@
       "./scripts/serializers/buffer.js"
     ],
     "setupTestFrameworkScriptFile": "./scripts/test/jestSetup.js"
+  },
+  "dependencies": {
+    "flow": "^0.2.3"
```
dicarlo2: Remove

packages/neo-one-monitor/src/BrowserLogger.js
```diff
@@ -0,0 +1,59 @@
...
+              message: options.error.message,
+              stack: options.error.stack,
+              code:
+                (options.error: $FlowFixMe).code == null
+                  ? options.error.constructor.name
+                  : (options.error: $FlowFixMe).code,
+            },
+    };
+
+    this._logs.push(log);
```
dicarlo2: nit: no need for a separate `log` variable, just do
```
this._logs.push({
  ...
})
```

packages/neo-one-monitor/src/BrowserMetricsFactory.js
```diff
@@ -0,0 +1,149 @@
...
+  // eslint-disable-next-line
+  observe(labels?: MetricLabels, count?: number): void {}
+}
+
+export default class BrowserMetricsFactory implements MetricsFactory {
+  collect(): MetricCollection {
+    const currentMetrics = this._serializeJSON(metrics);
+    utils
+      .keys(metrics)
+      .map(metricType =>
```
dicarlo2: This would be an appropriate time to use `forEach` since you're doing a side effect.

packages/neo-one-monitor/src/BrowserMetricsFactory.js
```diff
@@ -0,0 +1,149 @@
...
+
+export class BrowserSummary extends CollectingMetric implements SummaryMetric {
+  // eslint-disable-next-line
+  observe(labels?: MetricLabels, count?: number): void {}
+}
+
+export default class BrowserMetricsFactory implements MetricsFactory {
+  collect(): MetricCollection {
+    const currentMetrics = this._serializeJSON(metrics);
+    utils
```
dicarlo2: Use `utils.values`, e.g.
```
utils.values(metrics).forEach((innerMetrics) => 
  utils.values(innerMetrics).forEach((metric) => metric.reset())
);
```

packages/neo-one-monitor/src/BrowserMetricsFactory.js
```diff
@@ -0,0 +1,149 @@
...
+
+    return histogram;
+  }
+
+  createSummary(options: MetricConstruct): SummaryMetric {
+    return new BrowserSummary(options);
+  }
+
+  _serializeJSON(currentMetrics: MetricCollection): MetricCollection {
+    const copy = utils
```
dicarlo2: This should be something like
```
return utils.entries(currentMetrics).reduce(
  (acc, [metricType, metricTypeMetrics]) => ({
    ...acc,
    [metricType]: utils.entries(metricTypeMetrics).reduce(
      (innerAcc, [name, metric]) => ({
        ...innerAcc
        [name]: ...
      }),
      {},
    ),
  }),
  {},
);
```

But anyways, I'm not sure what the point of this is anyways, why are we doing a deep clone of `MetricCollection`?

dicarlo2: Oh I think I see now, you're converting all of the class instances to just the objects. I would do something like `[name]: metric.toJSON()` in the above example code to make that clearer and do the actual conversion inside the `Metric` class. It makes more sense from an OO perspective anyways, why should `MetricsFactory` know how to serialize a particular `Metric`? Instead, the `Metric` should know how to serialize itself.

packages/neo-one-monitor/src/types.js
```diff
@@ -197,6 +198,24 @@ export interface Summary {
   observe(labels: Labels, value: number): void;
 }
 
+export interface CounterMetric {
```
dicarlo2: These should stay in `MonitorBase` - they don't have anything to do with the public api of `Monitor` (which is what's in this file).

packages/neo-one-monitor/src/types.js
```diff
@@ -1,5 +1,6 @@
 /* @flow */
 import type { Context } from 'koa';
+import type { Report } from './Reporter';
```
dicarlo2: On the other hand, the definition of `Report` belongs in this file. `Report` is part of the public api of `Monitor`.

dicarlo2: One way to think about where things belong is to consider who should "know" about a particular type or class. `Monitor` doesn't need to know about `GaugeMetric`, etc. but it does need to "know" about `Report`, so it belong in this file.

packages/neo-one-monitor/src/MonitorBase.js
```diff
@@ -28,9 +33,17 @@ import type {
   SpanContext,
   SpanOptions,
   Summary,
+  SummaryMetric,
 } from './types';
+import type { Report } from './Reporter';
 import createTracer from './createTracer';
 
+export type FullLogLevelOption = {|
```
dicarlo2: Revert

packages/neo-one-monitor/src/MonitorBase.js
```diff
@@ -780,6 +810,66 @@ export default class MonitorBase implements Span {
...
+        labels: log.labels,
+        data: log.data,
+        error: error == null ? undefined : errorObj,
+      };
+
+      this._logger.log(newLog);
+    });
+
+    utils
+      .keys(report.metrics.counters)
```
dicarlo2: `.entries`

packages/neo-one-monitor/src/MonitorBase.js
```diff
@@ -780,6 +810,66 @@ export default class MonitorBase implements Span {
...
+      .map(counterName => {
```
dicarlo2: I would just do this in one `forEach`.

packages/neo-one-monitor/src/Reporter.js
```diff
@@ -0,0 +1,153 @@
...
+    this._subscription = interval(timer)
+      .pipe(
+        mergeScan(
+          async backReport => {
+            const logs = logger.collect();
+            const metrics = metricsFactory.collect();
+
+            const report = this._addBackReport({ logs, metrics }, backReport);
+
+            const response = await this._report(report);
```
dicarlo2: Need to catch errors here.

packages/neo-one-monitor/src/Reporter.js
```diff
@@ -0,0 +1,153 @@
...
+
+    const newMetrics = utils
+      .keys(backReport.metrics)
+      .reduce((accMetricType, metricType) => {
+        const resultMetricType = { ...accMetricType };
+        resultMetricType[metricType] = utils
+          .keys(backReport.metrics[metricType])
+          .reduce(
+            (accMetric, name) => {
+              const resultMetric = { ...accMetric };
```
dicarlo2: Same comments as above. Anywhere you're doing this pattern:
```
const copy = { ...original };
copy[key] = value;
```
Instead should be
```
const copy = {
   ...original
   [key]: value
};
```
Or just return the value directly if that's the only thing that's happening.

packages/neo-one-monitor/src/Reporter.js
```diff
@@ -0,0 +1,153 @@
...
+
+  _addBackReport(report: Report, backReport?: Report): Report {
+    if (backReport == null) {
+      return report;
+    }
+    const logs = backReport.logs.concat(report.logs);
+    const { metrics } = report;
+
+    const newMetrics = utils
+      .keys(backReport.metrics)
```
dicarlo2: Same comment here. Anywhere you're doing
```
utils.keys(obj).forEach((key) => {
  const value = obj[key];
  ...
})
```
Replace with
```
utils.entries(obj).forEach(([key, value]) => {
  ...
})
```

packages/neo-one-monitor/src/Reporter.js
```diff
@@ -0,0 +1,153 @@
...
+              return resultMetric;
+            },
+            { ...metrics[metricType] },
+          );
+        return resultMetricType;
+      }, {});
+
+    return {
+      logs,
+      metrics: {
```
dicarlo2: You don't need to copy them into yet another object. Just return `logs` and `newMetrics` directly.

packages/neo-one-monitor/src/Reporter.js
```diff
@@ -0,0 +1,153 @@
...
+      },
+    };
+  }
+
+  _constructBackReport(report: Report): Report {
+    const backLogs = report.logs.slice(-MAX_BACKLOG);
+
+    const backMetrics = utils
+      .keys(report.metrics)
+      .reduce((accMetricType, metricType) => {
```
dicarlo2: Same comments here about `utils.entries` and object cloning.

packages/neo-one-monitor/src/Reporter.js
```diff
@@ -0,0 +1,153 @@
...
+        const resultMetricType = { ...accMetricType };
+        resultMetricType[metricType] = utils
+          .keys(report.metrics[metricType])
+          .reduce((accMetric, name) => {
+            const result = { ...accMetric };
+            result[name] = report.metrics[metricType][name];
```
dicarlo2: We need to slice something here right? Otherwise this is just cloning the input metrics. Probably a case where you should be using `forEach` since you need to modify the metric values but not actually change the overall structure of `report.metrics`.

packages/neo-one-monitor/src/BrowserMetricsFactory.js
```diff
@@ -0,0 +1,151 @@
...
+  metric: MetricConstruct;
+  values: Array<MetricValue>;
+
+  constructor(metric: MetricConstruct) {
+    this.metric = metric;
+    this.values = [];
+  }
+
+  toJSON(): CollectingMetricJSON {
+    return {
```
dicarlo2: I don't think you need to copy anything here. Just
```
return { metric: this.metric, values: this.values }
```

packages/neo-one-monitor/src/BrowserMetricsFactory.js
```diff
@@ -0,0 +1,151 @@
...
+
+    return histogram;
+  }
+
+  createSummary(options: MetricConstruct): SummaryMetric {
+    return new BrowserSummary(options);
+  }
+
+  _serializeJSON(currentMetrics: MetricCollection): MetricCollection {
+    const copy = utils.entries(currentMetrics).reduce(
```
dicarlo2: `return ...`

packages/neo-one-monitor/src/BrowserMetricsFactory.js
```diff
@@ -0,0 +1,151 @@
...
+            ...accMetric,
+            [name]: metric.toJSON(),
+          }),
+          {},
+        ),
+      }),
+      { counters: {}, histograms: {} },
+    );
+
+    return {
```
dicarlo2: Remove

packages/neo-one-monitor/src/Reporter.js
```diff
@@ -0,0 +1,165 @@
...
+    timer: number,
+    endpoint: string,
+  |}) {
+    this._endpoint = endpoint;
+    this._requestCounter = metricsFactory.createCounter({
+      name: 'browser_request_counter',
+      help: 'Counts fetch requests from Browser Reporter to endpoint',
+      labelNames: [KNOWN_LABELS.ERROR],
+    });
+    const emptyBackReport = {
```
dicarlo2: nit: make this a constant at the module scope so you don't have to pass it in to `collectReport`

packages/neo-one-monitor/src/Reporter.js
```diff
@@ -0,0 +1,165 @@
...
+    });
+  }
+
+  async _collectReport({
+    backReport,
+    logger,
+    metricsFactory,
+    emptyBackReport,
+  }: {|
+    backReport?: Report,
```
dicarlo2: This should always be non-null.

packages/neo-one-monitor/src/Reporter.js
```diff
@@ -0,0 +1,165 @@
...
+
+    if (!response.ok) {
+      this._requestCounter.inc({ [KNOWN_LABELS.ERROR]: true });
+      return this._constructBackReport(report);
+    }
+    this._requestCounter.inc({ [KNOWN_LABELS.ERROR]: false });
+    return emptyBackReport;
+  }
+
+  _addBackReport(report: Report, backReport?: Report): Report {
```
dicarlo2: Same here, `backReport` is always non-null.

packages/neo-one-monitor/src/Reporter.js
```diff
@@ -0,0 +1,165 @@
...
+                  ? metric.values
+                  : metric.values.concat(
+                      currentMetrics[metricType][name].values,
+                    ),
+            },
+          }),
+          { ...currentMetrics[metricType] },
+        ),
+      }),
+      { counters: {}, histograms: {} },
```
dicarlo2: I think you want to set this to `currentMetrics`. Otherwise, if there are no metrics in `backReport.metrics` then `newMetrics` will always be this empty object rather than whatever is in `currentMetrics`.

So, change this line to just be `currentMetrics` (or just `report.metrics`) and change line 124 to be just `accMetricType[metricType] || {}`.

Also, a general note - remember, everything is treated as immutable, you don't need to copy everywhere. To be specific, we're always making copies on write (i.e. that's why it's treated as immutable), so we don't need to also make copies on read, e.g. what you're doing on line 124.

packages/neo-one-monitor/src/types.js
```diff
@@ -1,5 +1,7 @@
 /* @flow */
 import type { Context } from 'koa';
+import type { MetricCollection } from './BrowserMetricsFactory';
```
dicarlo2: Same thing here, if `MetricCollection` and `CollectingLoggerLogOptions` are part of the public api of `Monitor`, then they should be defined in this file, not somewhere else.

dicarlo2: I.e. this file should not import from any other file.

Issue #160, opened 2018-05-23
Handle circular references in ArrayContractParameter.


dicarlo2:
Title. Should probably fail fast when it sees a reference it's already seen. See ArrayContractParameter.js

davemneo:
Clarification exchanged via email:
in relation to ArrayLikeStackItem.toContractParameter. You should pass in a set of seen references, check if this is already in the seen set, if it is, throw an error, otherwise copy the seen set and add this to it. Similarly for MapStackItem, and any other stack items that could have circular references (i.e. they call toContractParameter recursively)
https://github.com/neo-one-suite/neo-one/blob/master/packages/neo-one-node-vm/src/stackItem/ArrayLikeStackItem.js

davemneo:
Submitted pull request for peer-review so we may discuss this possible solution if this path is reasonable:
https://github.com/neo-one-suite/neo-one/pull/187

davemneo:
see pull request: https://github.com/neo-one-suite/neo-one/pull/187

PR #186, opened 2018-06-01
handle circular references in ArrayContractParameter

davemneo:
addressing #160 

Sanity, correctness and style check

this commit was done with --no-verify

## Do Not Merge


davemneo:
Upon discussion with story author, we're withdrawing this pull request in favor of a similar but different one.

PR #187, opened 2018-06-04
Circular Reference check for Arrays toContractParameter

davemneo:
Resolves #187 
##errors.js:
added CircularReferenceError

## ArrayLikeStackItem.js
Updated toContractParameter with <code>seen</code> and supporting logic to track parent chain and test if current object is already in parent chain.

## For StackItemBase and all classes extending StackItemBase
updated .toContractParameter() to include parameter 'seen' & added eslint-disable to allow eslint to not throw UNUSED_VAR alarm.


davemneo:
Using a Set() does look like the preferred way to go.

Quick comparison of loading and accessing Set() and Object values: 
https://gist.github.com/davemneo/5939769f037b3f71da99b2f9c64e6f84

Note: the pre-commit verify has 10 errors:  toContractParameter calls do not expect arguments.
No change when wrapped this in an instanceof test.

 if(this instanceof ArrayLikeStackItem){
      return new ArrayContractParameter(this.value.map((value) => value.toContractParameter(newSeen)));
    }
    return new ArrayContractParameter(this.value.map((value) => value.toContractParameter()));

Using a slightly modified version of the recommended solution to avoid runtime errors.

packages/neo-one-node-vm/src/stackItem/ArrayLikeStackItem.js
```diff
@@ -55,10 +59,13 @@ export default class ArrayLikeStackItem extends StackItemBase {
     throw new InvalidValueBufferError();
   }
 
-  toContractParameter(): ContractParameter {
-    return new ArrayContractParameter(
-      this.value.map((val) => val.toContractParameter()),
-    );
+  toContractParameter(seen: Set<StackItemBase> = new Set()): ContractParameter {
```
dicarlo2: the interface in the base class needs to be updated to include this extra argument.

davemneo:
As we discussed, it's easier to do a fresh pull request with the updates after the conversion from TS to JS

PR #207, opened 2018-06-18
Circular Reference check for Array toParameters

### LINKED ISSUE #160: Handle circular references in ArrayContractParameter.

davemneo:
After migrating from JS to TS, we've dropped the previous pull request and started over.
https://github.com/neo-one-suite/neo-one/pull/187

Resolves #160
added CircularReferenceError

Updated toContractParameter with <code>seen</code> and supporting logic to track parent chain and test if current object is already in parent chain.

updated .toContractParameter() to include parameter 'seen' & added tslint-disable to allow tslint to not throw UNUSED_VAR alarm.


packages/neo-one-node-vm/src/stackItem/AttributeStackItem.ts
```diff
@@ -53,7 +54,8 @@ export class AttributeStackItem extends ObjectStackItem<Attribute> {
     return new BufferStackItem(value.value);
   }
 
-  public toContractParameter(): ContractParameter {
+  // tslint:disable-next-line no-unused
+  public toContractParameter(seen: Set<StackItemBase> = new Set()): ContractParameter {
```
dicarlo2: Use `_seen` and remove the `no-unused` disable

dicarlo2: Here and throughout the PR

packages/neo-one-node-vm/src/stackItem/ArrayLikeStackItem.ts
```diff
@@ -50,8 +50,15 @@ export class ArrayLikeStackItem extends StackItemBase {
     throw new InvalidValueBufferError();
   }
 
-  public toContractParameter(): ContractParameter {
-    return new ArrayContractParameter(this.value.map((val) => val.toContractParameter()));
+  // tslint:disable-next-line no-unused
```
dicarlo2: Remove

PR #222, opened 2018-06-19
Export BrowserMonitor

afragapane:
### Description of the Change
Just adds the export of BrowserMonitor for access externally.


dicarlo2:
This is not necessary.

PR #224, opened 2018-06-20
Keep track of metrics in MetricsFactoryProxy

afragapane:
### Description of the Change
Previously, ReportHandler kept track of which metrics it had forwarded to the NodeMetricsFactory.  This caused issues where in some cases the NodeMetricsFactory had already registered some of those metrics with prometheus.  Now, the MetricsFactoryProxy keeps track of all metrics which have been registered, so there should be no issue when forwarding metrics from the browser.

### DIFF
--- a/packages/neo-one-monitor/src/MetricsFactoryProxy.ts
+++ b/packages/neo-one-monitor/src/MetricsFactoryProxy.ts
@@ -93,38 +93,58 @@ export class SummaryProxy extends MetricProxy<SummaryBase> implements Summary {
 }
 
 export class MetricsFactoryProxy implements MetricsFactory {
+  protected mutableCounters: { [K in string]?: Counter } = {};
+  protected mutableHistograms: { [K in string]?: Histogram } = {};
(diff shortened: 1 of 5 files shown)


packages/neo-one-monitor/src/MetricsFactoryProxy.ts
```diff
@@ -93,38 +93,56 @@ export class SummaryProxy extends MetricProxy<SummaryBase> implements Summary {
...
+  protected mutableCounters: { [K in string]?: Counter } = {};
+  protected mutableHistograms: { [K in string]?: Histogram } = {};
+  protected mutableGauges: { [K in string]?: Gauge } = {};
+  protected mutableSummaries: { [K in string]?: Summary } = {};
+
   private mutableFactory: MetricsFactory | undefined;
 
+  public reset_forTest(): void {
+    this.mutableCounters = {};
+    this.mutableHistograms = {};
```
dicarlo2: Probably also want to reset `mutableGauges` and `mutableSummaries`

Issue #43, opened 2018-01-02
Add argument checking assertions to @neo-one/client Client


dicarlo2:
### Description

Validate all arguments provided to Client methods.

dicarlo2:
E.g. https://github.com/neo-one-suite/neo-one/blob/master/packages/neo-one-client/src/Client.js#L65

danwbyrne:
Your link from Feb 10th is no longer valid. If this is a priority card let me know. I've currently got multiple inputs & consolidating inputs as my priority.

danwbyrne:
https://github.com/danwbyrne/neo-one/tree/client-arg-checking

Possible bug on an already defined method "assertBuffer".

when checking with: 
```typescript
if (Buffer.from(value, 'hex').toString('hex') ...
```
a character will be chopped off of odd-length hex strings.

I have at least one test per new assert method right now and all are passing. More to be done though.

PR #232, opened 2018-06-21
Adds client argument assertion checks

danwbyrne:
Adds client argument assertion checks.

Wrote and implement additional assertions for ~all arguments of public Client.ts methods. Wrote additional pass/fail tests in args.test.ts, as well as updating tests in Client.test.ts because they were built on the premise that they can pass anything into argument fields. 

Passing all tests currently.

Addresses issue #43

### DIFF
--- a/packages/neo-one-client/src/Client.ts
+++ b/packages/neo-one-client/src/Client.ts
@@ -39,7 +39,6 @@ export class Client<TUserAccountProviders extends { readonly [K in string]: User
   public static inject(provider: UserAccountProvider): void {
     mutableClients.forEach((client) => client.inject(provider));
   }
-
   public readonly currentAccount$: Observable<UserAccount | undefined>;
   public readonly accounts$: Observable<ReadonlyArray<UserAccount>>;
   public readonly networks$: Observable<ReadonlyArray<NetworkType>>;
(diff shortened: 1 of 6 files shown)


packages/neo-one-client/src/Client.ts
```diff
@@ -145,6 +149,7 @@ export class Client<TUserAccountProviders extends { readonly [K in string]: User
     return [...new Set(providers.reduce((acc: NetworkType[], provider) => acc.concat(provider.getNetworks()), []))];
   }
 
+  // tslint:disable-next-line readonly-array no-any
```
dicarlo2: Remove

packages/neo-one-client/src/Client.ts
```diff
@@ -181,19 +187,12 @@ export class Client<TUserAccountProviders extends { readonly [K in string]: User
     asset: AssetRegister,
     options?: TransactionOptions,
   ): Promise<TransactionResult<RegisterAssetReceipt>> {
+    argAssertions.assertAssetRegister(asset);
+    argAssertions.assertTransactionOptions(options);
+
     return this.getProvider(options).registerAsset(asset, options);
   }
 
-  public async issue(
```
dicarlo2: Revert

packages/neo-one-client/src/Client.ts
danwbyrne: The merge conflict was because of a source map argument getting added to a call that hadn't been rebased because I hadn't worked on it. I believe everything should be fixed now.

Issue #221, opened 2018-06-19
Add a Client method for consolidating outputs


dicarlo2:
Ledger has a maximum transaction size that it can support, which means that if an address has a large number of small outputs it may be impossible to construct a single transaction within that limit. Let's add a method to the [`Client`](https://github.com/neo-one-suite/neo-one/blob/master/packages/neo-one-client/src/Client.ts#L34) with the following signature:

```
public consolidate(asset: Hash256String, maxInputs: number, options?: TransactionOptions): Promise<TransactionResult<TransactionReceipt>>;
```

that will create a contract transaction (similar to transfer) that will consolidate the smallest `maxInputs` inputs into a single output to the same address.

danwbyrne:
After talking about this problem with Frag, and just so I'm clear on how this should be executed I think the plan of attack here is to use the UserAccount address to get the unspent input/outputs for that address, and then consolidate those outputs through some process.

PR #272, opened 2018-06-28
Adds Unspent Output Consolidation #221

danwbyrne:
Working currently in tests. Consolidates unspent outputs when a byte-limit is specified on the keystore to avoid running into the problem of not being able to make a large transaction with multiple small inputs because of ledger/wallet restrictions.

I'm pretty confident in the process and have a test for contractTransaction and claimTransaction that both pass with / without network fees and variable byteLimits.

edit: fixed red squiggles.

### DIFF
--- a/packages/neo-one-client-core/src/transaction/Input.ts
+++ b/packages/neo-one-client-core/src/transaction/Input.ts
@@ -21,7 +21,9 @@ export interface InputJSON {
   readonly vout: number;
 }
 
+const SIZE = IOHelper.sizeOfUInt256 + IOHelper.sizeOfUInt16LE;
 export class Input implements SerializableWire<Input>, Equatable, SerializableJSON<InputJSON> {
+  public static readonly size: number = SIZE;
   public static deserializeWireBase({ reader }: DeserializeWireBaseOptions): Input {
     const hash = reader.readUInt256();
     const index = reader.readUInt16LE();
@@ -38,7 +40,7 @@ export class Input implements SerializableWire<Input>, Equatable, SerializableJS
 
   public readonly hash: UInt256;
   public readonly index: number;
-  public readonly size: number = IOHelper.sizeOfUInt256 + IOHelper.sizeOfUInt16LE;
+  public readonly size: number = SIZE;
   public readonly equals: Equals = utils.equals(
     Input,
     (other) => common.uInt256Equal(this.hash, other.hash) && other.index === this.index,
--- a/packages/neo-one-client/src/__tests__/user/LocalUserAccountProvider.test.ts
+++ b/packages/neo-one-client/src/__tests__/user/LocalUserAccountProvider.test.ts
@@ -1,5 +1,5 @@
 // tslint:disable
-import { common, utils } from '@neo-one/client-core';
+import { common, utils, Transaction } from '@neo-one/client-core';
 import BigNumber from 'bignumber.js';
 import {
   InsufficientFundsError,
@@ -39,28 +39,28 @@ describe('LocalUserAccountProvider', () => {
       asset: common.NEO_ASSET_HASH,
       value: new BigNumber('1'),
       address: 'ALq7AWrhAueN6mJNqk6FHJjnsEoPRytLdW',
-      txid: '0x7f48028c38117ac9e42c8e1f6f06ae027cdbb904eaf1a0bdc30c9d81694e045c',
+      txid: '0x7f48028c38117ac9e42c8e1f6f06ae027cdbb904eaf1a0bdc30c9d81694e045a',
       vout: 1,
     },
     twoNEO: {
       asset: common.NEO_ASSET_HASH,
-      value: new BigNumber('1'),
+      value: new BigNumber('2'),
       address: 'ALq7AWrhAueN6mJNqk6FHJjnsEoPRytLdW',
-      txid: '0x7f48028c38117ac9e42c8e1f6f06ae027cdbb904eaf1a0bdc30c9d81694e045c',
+      txid: '0x7f48028c38117ac9e42c8e1f6f06ae027cdbb904eaf1a0bdc30c9d81694e045a',
       vout: 2,
     },
     sevenNEO: {
       asset: common.NEO_ASSET_HASH,
       value: new BigNumber('7'),
       address: 'ALq7AWrhAueN6mJNqk6FHJjnsEoPRytLdW',
-      txid: '0x7f48028c38117ac9e42c8e1f6f06ae027cdbb904eaf1a0bdc30c9d81694e045d',
+      txid: '0x7f48028c38117ac9e42c8e1f6f06ae027cdbb904eaf1a0bdc30c9d81694e045a',
       vout: 3,
     },
     oneGAS: {
       asset: common.GAS_ASSET_HASH,
-      value: new BigNumber('2'),
+      value: new BigNumber('1'),
       address: 'ALq7AWrhAueN6mJNqk6FHJjnsEoPRytLdW',
-      txid: '0x7f48028c38117ac9e42c8e1f6f06ae027cdbb904eaf1a0bdc30c9d81694e045e',
+      txid: '0x7f48028c38117ac9e42c8e1f6f06ae027cdbb904eaf1a0bdc30c9d81694e045d',
       vout: 4,
     },
     twoGAS: {
@@ -76,6 +76,27 @@ describe('LocalUserAccountProvider', () => {
       address: 'ALq7AWrhAueN6mJNqk6FHJjnsEoPRytLdW',
       txid: '0x7f48028c38117ac9e42c8e1f6f06ae027cdbb904eaf1a0bdc30c9d81694e045e',
       vout: 6,
+    },
+    oneTKY: {
+      asset: '0x7f48028c38117ac9e42c8e1f6f06ae027cdbb904eaf1a0bdc30c9d81694e045c',
+      value: new BigNumber('1'),
+      address: 'ALq7AWrhAueN6mJNqk6FHJjnsEoPRytLdW',
+      txid: '0x7f48028c38117ac9e42c8e1f6f06ae027cdbb904eaf1a0bdc30c9d81694e045c',
+      vout: 7,
+    },
+    twoTKY: {
+      asset: '0x7f48028c38117ac9e42c8e1f6f06ae027cdbb904eaf1a0bdc30c9d81694e045c',
+      value: new BigNumber('2'),
+      address: 'ALq7AWrhAueN6mJNqk6FHJjnsEoPRytLdW',
+      txid: '0x7f48028c38117ac9e42c8e1f6f06ae027cdbb904eaf1a0bdc30c9d81694e045c',
+      vout: 8,
+    },
+    threeTKY: {
+      asset: '0x7f48028c38117ac9e42c8e1f6f06ae027cdbb904eaf1a0bdc30c9d81694e045c',
+      value: new BigNumber('3'),
+      address: 'ALq7AWrhAueN6mJNqk6FHJjnsEoPRytLdW',
+      txid: '0x7f48028c38117ac9e42c8e1f6f06ae027cdbb904eaf1a0bdc30c9d81694e045c',
+      vout: 9,
     },
   };
 
@@ -1059,7 +1080,6 @@ describe('LocalUserAccountProvider', () => {
     const result1 = await localUserAccountProvider.transfer(transfers1, {
       from: options.from,
       attributes: options.attributes,
-      networkFee: undefined,
     });
 
     provider.getBlockCount = jest.fn(() => Promise.resolve(1));
@@ -1067,13 +1087,207 @@ describe('LocalUserAccountProvider', () => {
     const result2 = await localUserAccountProvider.transfer(transfers2, {
       from: options.from,
       attributes: options.attributes,
-      networkFee: undefined,
     });
 
     await result1;
     // @ts-ignore
     expect(localUserAccountProvider.mutableUsedOutputs.size).toEqual(1);
     expect(result2).toHaveProperty('transaction');
+    verifyMocks();
+  });
+
+  test('transaction consolidates', async () => {
+    const unspent = [outputs.oneNEO, outputs.twoNEO, outputs.sevenNEO, outputs.oneGAS, outputs.elevenGAS];
+    const transfer = [
+      {
+        amount: new BigNumber('3'),
+        asset: common.NEO_ASSET_HASH,
+        to: 'AVf4UGKevVrMR1j3UkPsuoYKSC4ocoAkKx',
+      },
+    ];
+    keystore.byteLimit = 800;
+    keystore.sign = jest.fn(() => witness);
+    provider.getUnspentOutputs = jest.fn(() => Promise.resolve(unspent));
+    provider.getTransactionReceipt = jest.fn(() => Promise.resolve('receiptDummy'));
+
+    const spy = jest.spyOn(localUserAccountProvider, 'addWitness' as any);
+    await localUserAccountProvider.transfer(transfer, {
+      from: options.from,
+      attributes: options.attributes,
+    });
+    const consolidatedTransaction = spy.mock.calls[0][0].transaction;
+    expect(consolidatedTransaction.serializeUnsigned().length).toEqual(366);
+    expect(consolidatedTransaction.inputs.length).toEqual(5);
+    expect(consolidatedTransaction.outputs.length).toEqual(3);
+
+    spy.mockReset();
+    spy.mockRestore();
+    verifyMocks();
+  });
+
+  test(`claim consolidates`, async () => {
+    const unspent = [outputs.elevenGAS];
+
+    keystore.byteLimit = 500;
+    keystore.sign = jest.fn(() => witness);
+    provider.getUnspentOutputs = jest.fn(() => Promise.resolve(unspent));
+    provider.getUnclaimed = jest.fn(() =>
+      Promise.resolve({
+        unclaimed: [
+          {
+            txid: '0x7f48028c36117ac9e42c8e1f6f06ae027cdbb904eaf1a0bdc30c9d81694e045c',
+            vout: 2,
+          },
+        ],
+
+        amount: new BigNumber('3'),
+      }),
+    );
+    const spy = jest.spyOn(localUserAccountProvider, 'addWitness' as any);
+
+    provider.getTransactionReceipt = jest.fn(() => Promise.resolve('receiptDummy'));
+
+    await localUserAccountProvider.claim({
+      from: options.from,
+      attributes: options.attributes,
+    });
+
+    const consolidatedTransaction = spy.mock.calls[0][0].transaction;
+    expect(consolidatedTransaction.serializeUnsigned().length).toEqual(145);
+    expect(consolidatedTransaction.inputs.length).toEqual(1);
+    expect(consolidatedTransaction.outputs.length).toEqual(1);
+
+    spy.mockReset();
+    spy.mockRestore();
+    verifyMocks();
+  });
+
+  test(`blank transaction consolidates`, async () => {
+    const unspent = [
+      outputs.oneNEO,
+      outputs.twoNEO,
+      outputs.sevenNEO,
+      outputs.oneGAS,
+      outputs.twoGAS,
+      outputs.elevenGAS,
+    ];
+    keystore.byteLimit = 800;
+    keystore.sign = jest.fn(() => witness);
+    provider.getUnspentOutputs = jest.fn(() => Promise.resolve(unspent));
+    provider.getTransactionReceipt = jest.fn(() => Promise.resolve('receiptDummy'));
+
+    const spy = jest.spyOn(localUserAccountProvider, 'addWitness' as any);
+    await localUserAccountProvider.transfer([], {
+      from: options.from,
+      attributes: options.attributes,
+    });
+    const consolidatedTransaction = spy.mock.calls[0][0].transaction;
+    expect(consolidatedTransaction.serializeUnsigned().length).toEqual(340);
+    expect(consolidatedTransaction.inputs.length).toEqual(6);
+    expect(consolidatedTransaction.outputs.length).toEqual(2);
+
+    spy.mockReset();
+    spy.mockRestore();
+    verifyMocks();
+  });
+
+  test(`consolidate prioritizes any asset with a return output`, async () => {
+    const unspent = [
+      outputs.oneTKY,
+      outputs.threeTKY,
+      outputs.oneNEO,
+      outputs.twoNEO,
+      outputs.sevenNEO,
+      outputs.oneGAS,
+      outputs.twoGAS,
+      outputs.elevenGAS,
+    ];
+
+    const transfers = [
+      {
+        amount: new BigNumber('2'),
+        asset: '0x7f48028c38117ac9e42c8e1f6f06ae027cdbb904eaf1a0bdc30c9d81694e045c',
+        to: 'AVf4UGKevVrMR1j3UkPsuoYKSC4ocoAkKx',
+      },
+    ];
+
+    keystore.byteLimit = 250;
+    keystore.sign = jest.fn(() => witness);
+    provider.getUnspentOutputs = jest.fn(() => Promise.resolve(unspent));
+    provider.getTransactionReceipt = jest.fn(() => Promise.resolve('receiptDummy'));
+
+    const spy = jest.spyOn(localUserAccountProvider, 'addWitness' as any);
+    await localUserAccountProvider.transfer(transfers, {
+      from: options.from,
+      attributes: options.attributes,
+    });
+    const consolidatedTransaction = spy.mock.calls[0][0].transaction as Transaction;
+    expect(consolidatedTransaction.serializeUnsigned().length).toEqual(204);
+    expect(
+      consolidatedTransaction.inputs.every((input) => common.uInt256ToString(input.hash) === outputs.oneTKY.asset),
+    ).toBe(true);
+
+    spy.mockReset();
+    spy.mockRestore();
+    verifyMocks();
+  });
+
+  test(`consolidation prioritizes NEO/GAS for new outputs`, async () => {
+    const unspent = [
+      outputs.oneTKY,
+      outputs.twoTKY,
+      outputs.oneNEO,
+      outputs.twoNEO,
+      outputs.sevenNEO,
+      outputs.oneGAS,
+      outputs.twoGAS,
+      outputs.elevenGAS,
+    ];
+
+    keystore.byteLimit = 400;
+    keystore.sign = jest.fn(() => witness);
+    provider.getUnspentOutputs = jest.fn(() => Promise.resolve(unspent));
+    provider.getTransactionReceipt = jest.fn(() => Promise.resolve('receiptDummy'));
+
+    const spy = jest.spyOn(localUserAccountProvider, 'addWitness' as any);
+    await localUserAccountProvider.transfer([], {
+      from: options.from,
+      attributes: options.attributes,
+    });
+    const consolidatedTransaction = spy.mock.calls[0][0].transaction as Transaction;
+    expect(consolidatedTransaction.inputs.some((input) => input.index === 7 || input.index === 8)).toBe(false);
+
+    spy.mockReset();
+    spy.mockRestore();
+    verifyMocks();
+  });
+
+  test('transaction consolidates GAS when network fee present', async () => {
+    const unspent = [outputs.oneNEO, outputs.oneGAS, outputs.elevenGAS];
+    const transfer = [
+      {
+        amount: new BigNumber('1'),
+        asset: common.NEO_ASSET_HASH,
+        to: 'AVf4UGKevVrMR1j3UkPsuoYKSC4ocoAkKx',
+      },
+    ];
+    keystore.byteLimit = 800;
+    keystore.sign = jest.fn(() => witness);
+    provider.getUnspentOutputs = jest.fn(() => Promise.resolve(unspent));
+    provider.getTransactionReceipt = jest.fn(() => Promise.resolve('receiptDummy'));
+
+    const spy = jest.spyOn(localUserAccountProvider, 'addWitness' as any);
+    await localUserAccountProvider.transfer(transfer, {
+      from: options.from,
+      attributes: options.attributes,
+      networkFee: new BigNumber('2'),
+    });
+    const consolidatedTransaction = spy.mock.calls[0][0].transaction;
+    expect(consolidatedTransaction.inputs.length).toEqual(3);
+    expect(consolidatedTransaction.outputs.length).toEqual(2);
+
+    spy.mockReset();
+    spy.mockRestore();
     verifyMocks();
   });
 });
--- a/packages/neo-one-client/src/user/LocalUserAccountProvider.ts
+++ b/packages/neo-one-client/src/user/LocalUserAccountProvider.ts
@@ -44,6 +44,7 @@ import {
   AssetRegister,
   Attribute,
   AttributeArg,
+  BufferString,
   ContractRegister,
   DataProvider,
   GetOptions,
@@ -76,13 +77,13 @@ import {
   UserAccountID,
   UserAccountProvider,
   Witness,
-  BufferString,
 } from '../types';
 import * as clientUtils from '../utils';
 import { converters } from './converters';
 
 export interface KeyStore {
   readonly type: string;
+  readonly byteLimit?: number;
   readonly currentAccount$: Observable<UserAccount | undefined>;
   readonly getCurrentAccount: () => UserAccount | undefined;
   readonly accounts$: Observable<ReadonlyArray<UserAccount>>;
@@ -218,7 +219,7 @@ export class LocalUserAccountProvider<TKeyStore extends KeyStore, TProvider exte
           monitor: span,
         });
 
-        if (inputs.length === 0) {
+        if (inputs.length === 0 && this.keystore.byteLimit === undefined) {
           throw new NothingToTransferError();
         }
 
@@ -231,6 +232,7 @@ export class LocalUserAccountProvider<TKeyStore extends KeyStore, TProvider exte
         return this.sendTransaction({
           from,
           transaction,
+          inputs,
           onConfirm: async ({ receipt }) => receipt,
           monitor: span,
         });
@@ -283,6 +285,7 @@ export class LocalUserAccountProvider<TKeyStore extends KeyStore, TProvider exte
         });
 
         return this.sendTransaction({
+          inputs,
           from,
           transaction,
           onConfirm: async ({ receipt }) => receipt,
@@ -451,6 +454,7 @@ export class LocalUserAccountProvider<TKeyStore extends KeyStore, TProvider exte
         });
 
         return this.sendTransaction({
+          inputs,
           from,
           transaction,
           onConfirm: async ({ receipt }) => receipt,
@@ -734,6 +738,7 @@ export class LocalUserAccountProvider<TKeyStore extends KeyStore, TProvider exte
 
         return this.sendTransaction({
           from,
+          inputs,
           transaction: invokeTransaction,
           onConfirm: async ({ transaction, receipt }) => {
             const data = await this.provider.getInvocationData(from.network, transaction.txid);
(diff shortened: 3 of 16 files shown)


packages/neo-one-client/src/user/LocalUserAccountProvider.ts
```diff
@@ -823,6 +841,172 @@ export class LocalUserAccountProvider<TKeyStore extends KeyStore, TProvider exte
...
+        value: output.value.add(clientUtils.bigNumberToBN(amount, 8)),
+        address: output.address,
+      });
+
+    const consoladatableUnspentsIn = await this.provider.getUnspentOutputs(from.network, from.address, monitor);
+    consoladatableUnspentsIn
+      .filter(
+        (unspent) =>
+          !inputs.some((input) => unspent.txid === input.txid && unspent.vout === input.vout) &&
+          !this.mutableUsedOutputs.has(`${unspent.txid}:${unspent.vout}`),
```
dicarlo2: This seems error-prone, instead, now every time we call `this.provider.getUnspentOutputs` we need to remember to check the `mutableUsedOutputs` set. Instead, you should make a private function `getUnspentOutputs` that does this the `mutableUsedOutputs` filtering.

packages/neo-one-client/src/user/LocalUserAccountProvider.ts
```diff
@@ -823,6 +841,172 @@ export class LocalUserAccountProvider<TKeyStore extends KeyStore, TProvider exte
...
+        asset: output.asset,
+        value: output.value.add(clientUtils.bigNumberToBN(amount, 8)),
+        address: output.address,
+      });
+
+    const consoladatableUnspentsIn = await this.provider.getUnspentOutputs(from.network, from.address, monitor);
+    consoladatableUnspentsIn
+      .filter(
+        (unspent) =>
+          !inputs.some((input) => unspent.txid === input.txid && unspent.vout === input.vout) &&
```
dicarlo2: This has complexity of O(n^2). Stick them in a set similar to `mutableUsedOutputs` and filter that way.

packages/neo-one-client/src/user/LocalUserAccountProvider.ts
```diff
@@ -823,6 +841,172 @@ export class LocalUserAccountProvider<TKeyStore extends KeyStore, TProvider exte
...
+          !this.mutableUsedOutputs.has(`${unspent.txid}:${unspent.vout}`),
+      )
+      .forEach((unspent) => {
```
dicarlo2: Use `reduce` and eliminate the local `let looseAssets` variable.

packages/neo-one-client/src/user/LocalUserAccountProvider.ts
```diff
@@ -823,6 +841,172 @@ export class LocalUserAccountProvider<TKeyStore extends KeyStore, TProvider exte
...
+        if ((looseAssets[unspent.asset] as ReadonlyArray<UnspentOutput> | undefined) !== undefined) {
```
dicarlo2: nit: typically easier to read the positive `if` condition first, i.e. `=== undefined`

packages/neo-one-client/src/user/LocalUserAccountProvider.ts
```diff
@@ -823,6 +841,172 @@ export class LocalUserAccountProvider<TKeyStore extends KeyStore, TProvider exte
...
+          };
+        }
+      });
+
+    const returnOutputs = transactionUnsignedIn.outputs.filter(
+      (output) => common.uInt160ToString(output.address) === addressToScriptHash(from.address),
+    );
+    let newUnspents: ReadonlyArray<UnspentOutput>;
+
+    if (returnOutputs.length > 0) {
```
dicarlo2: If `returnOutputs.length === 0` the following lines are a no-op, so you don't need to check that it's larger.

packages/neo-one-client/src/user/LocalUserAccountProvider.ts
```diff
@@ -823,6 +841,172 @@ export class LocalUserAccountProvider<TKeyStore extends KeyStore, TProvider exte
...
+            undefined,
+        )
+        .forEach((output) => {
+          const currentAsset = common.uInt256ToString(output.asset);
+          const toAdd = Math.min(
+            looseAssets[currentAsset].length,
+            Math.floor((byteLimit - getMessageSize()) / newInputSize),
+          );
+          if (toAdd === 0) {
+            return createConsolidatedTransaction();
```
dicarlo2: ?? This does nothing.

packages/neo-one-client/src/user/LocalUserAccountProvider.ts
```diff
@@ -823,6 +841,172 @@ export class LocalUserAccountProvider<TKeyStore extends KeyStore, TProvider exte
...
+          updatedOutputs = updatedOutputs.map((out) => {
+            if (out.address === output.address && out.asset === output.asset) {
+              return updateOutput(out, newUnspents.reduce((left, right) => left.plus(right.value), new BigNumber('0')));
+            }
+
+            return out;
+          });
+          newInputs = newInputs.concat(this.convertInputs(newUnspents));
+
+          return;
```
dicarlo2: ??

packages/neo-one-client/src/user/LocalUserAccountProvider.ts
```diff
@@ -823,6 +841,172 @@ export class LocalUserAccountProvider<TKeyStore extends KeyStore, TProvider exte
...
+            );
+          const newOutput: Output = {
+            asset: key,
+            value: newUnspents.reduce((left, right) => left.plus(right.value), new BigNumber('0')),
+            address: from.address,
+          };
+          newOutputs = newOutputs.concat(this.convertOutputs([newOutput]));
+          newInputs = newInputs.concat(this.convertInputs(newUnspents));
+
+          return;
```
dicarlo2: ??

packages/neo-one-client/src/user/LocalUserAccountProvider.ts
```diff
@@ -823,6 +841,172 @@ export class LocalUserAccountProvider<TKeyStore extends KeyStore, TProvider exte
...
+        }
+
+        return createConsolidatedTransaction();
+      });
+
+    return createConsolidatedTransaction();
```
dicarlo2: This does something. But why a separate function? Just move that code down here.

packages/neo-one-client/src/__tests__/user/LocalUserAccountProvider.test.ts
```diff
@@ -1076,4 +1076,60 @@ describe('LocalUserAccountProvider', () => {
...
+    provider.getTransactionReceipt = jest.fn(() => Promise.resolve('receiptDummy'));
+
+    const spy = jest.spyOn(keystore, 'sign');
+    await localUserAccountProvider.transfer(transfer, {
+      from: options.from,
+      attributes: options.attributes,
+      networkFee: undefined,
+    });
+
+    expect(spy.mock.calls[0][0].message.length).toEqual(732);
```
dicarlo2: This still doesn't really show what's being verified. Sure, the size is less than 800, but don't we expect more inputs and outputs? You should be verifying that the transaction has the inputs and outputs you expect.

packages/neo-one-client/src/__tests__/user/LocalUserAccountProvider.test.ts
```diff
@@ -1076,4 +1076,60 @@ describe('LocalUserAccountProvider', () => {
...
+    const spy = jest.spyOn(keystore, 'sign');
+
+    provider.getTransactionReceipt = jest.fn(() => Promise.resolve('receiptDummy'));
+
+    await localUserAccountProvider.claim({
+      from: options.from,
+      attributes: options.attributes,
+      networkFee: undefined,
+    });
+    expect(spy.mock.calls[0][0].message.length).toEqual(290);
```
dicarlo2: Same here.

packages/neo-one-client/src/user/LocalUserAccountProvider.ts
```diff
@@ -823,6 +841,172 @@ export class LocalUserAccountProvider<TKeyStore extends KeyStore, TProvider exte
...
+    monitor,
+    byteLimit,
+  }: {
+    readonly transactionUnsignedIn: TransactionModel;
+    readonly inputs: ReadonlyArray<UnspentOutput>;
+    readonly from: UserAccountID;
+    readonly monitor?: Monitor;
+    readonly byteLimit: number;
+  }): Promise<TransactionModel> {
+    const newInputSize = 68;
```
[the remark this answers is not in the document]
dicarlo2: Still needs fixing

packages/neo-one-client/src/user/LocalUserAccountProvider.ts
```diff
@@ -823,6 +841,172 @@ export class LocalUserAccountProvider<TKeyStore extends KeyStore, TProvider exte
...
+    const newOutputSize = 120;
```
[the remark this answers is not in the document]
dicarlo2: Still needs fixing

packages/neo-one-client/src/user/LocalUserAccountProvider.ts
```diff
@@ -823,6 +841,172 @@ export class LocalUserAccountProvider<TKeyStore extends KeyStore, TProvider exte
...
+    const messageSize = transactionUnsignedIn.serializeUnsigned().toString('hex').length;
```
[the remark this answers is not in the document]
dicarlo2: Still needs fixing

packages/neo-one-client/src/user/LocalUserAccountProvider.ts
```diff
@@ -823,6 +841,172 @@ export class LocalUserAccountProvider<TKeyStore extends KeyStore, TProvider exte
...
+
+    if (byteLimit - messageSize < 0) {
+      throw new TransactionSizeExceedsLimitError(messageSize, byteLimit);
+    }
+    if (byteLimit - messageSize - newInputSize < 0) {
+      return transactionUnsignedIn;
+    }
+
+    const inputMap = (): { [key: string]: ReadonlyArray<UnspentOutput> } => ({});
```
dicarlo2: Remove. This is a really weird pattern.

packages/neo-one-client/src/user/LocalUserAccountProvider.ts
```diff
@@ -823,6 +841,172 @@ export class LocalUserAccountProvider<TKeyStore extends KeyStore, TProvider exte
...
+    const inputArray = (): ReadonlyArray<InputModel> => [];
```
dicarlo2: Remove.

packages/neo-one-client/src/user/LocalUserAccountProvider.ts
```diff
@@ -823,6 +841,172 @@ export class LocalUserAccountProvider<TKeyStore extends KeyStore, TProvider exte
...
+    const outputArray = (): ReadonlyArray<OutputModel> => [];
```
dicarlo2: Remove.

packages/neo-one-client/src/user/LocalUserAccountProvider.ts
```diff
@@ -823,6 +841,172 @@ export class LocalUserAccountProvider<TKeyStore extends KeyStore, TProvider exte
...
+    const getMessageSize = (numInputs: number, numOutputs: number) =>
+      messageSize + numInputs * newInputSize + numOutputs * newOutputSize;
```
dicarlo2: Change this to `messageSize + (numInputs * newInputSize) + (numOutputs * newOutputSize)` to make the order of operations clearer.

packages/neo-one-client/src/user/LocalUserAccountProvider.ts
```diff
@@ -823,6 +841,172 @@ export class LocalUserAccountProvider<TKeyStore extends KeyStore, TProvider exte
...
+
+    const updateOutput = (output: OutputModel, amount: BigNumber): OutputModel =>
```
dicarlo2: Move this to a method on `OutputModel` (`clone`).

packages/neo-one-client/src/user/LocalUserAccountProvider.ts
```diff
@@ -823,6 +841,172 @@ export class LocalUserAccountProvider<TKeyStore extends KeyStore, TProvider exte
...
+      new OutputModel({
+        asset: output.asset,
+        value: output.value.add(clientUtils.bigNumberToBN(amount, 8)),
+        address: output.address,
+      });
+
+    const consoladatableUnspentsIn = await this.provider.getUnspentOutputs(from.network, from.address, monitor);
```
dicarlo2: `consoladatableUnspents`

Also... `consolidatable` not `consoladatable`

packages/neo-one-client/src/user/LocalUserAccountProvider.ts
```diff
@@ -823,6 +841,172 @@ export class LocalUserAccountProvider<TKeyStore extends KeyStore, TProvider exte
...
+
+    const looseAssets = consoladatableUnspentsIn
+      .filter(
+        (unspent) =>
+          !inputs.some((input) => unspent.txid === input.txid && unspent.vout === input.vout) &&
+          !this.mutableUsedOutputs.has(`${unspent.txid}:${unspent.vout}`),
```
dicarlo2: Same comment as the previous review.

packages/neo-one-client/src/user/LocalUserAccountProvider.ts
```diff
@@ -823,6 +841,172 @@ export class LocalUserAccountProvider<TKeyStore extends KeyStore, TProvider exte
...
+      )
+      .reduce((acc, unspent) => {
+        if ((acc[unspent.asset] as ReadonlyArray<UnspentOutput> | undefined) !== undefined) {
```
dicarlo2: Same comment as previous review.

packages/neo-one-client/src/user/LocalUserAccountProvider.ts
```diff
@@ -823,6 +841,172 @@ export class LocalUserAccountProvider<TKeyStore extends KeyStore, TProvider exte
...
+            .sort((coinA, coinB) => coinA.value.comparedTo(coinB.value))
+            .slice(0, toAdd);
+          const newUpdatedOutputs = acc.updatedOutputs.map((out) => {
+            if (out.address === output.address && out.asset === output.asset) {
+              return updateOutput(out, tempIns.reduce((left, right) => left.plus(right.value), new BigNumber('0')));
+            }
+
+            return out;
+          });
+          const evenLooserAssets = {
```
dicarlo2: `newLooseAssets`

packages/neo-one-client/src/user/LocalUserAccountProvider.ts
```diff
@@ -823,6 +841,172 @@ export class LocalUserAccountProvider<TKeyStore extends KeyStore, TProvider exte
...
+    const updateOutput = (output: OutputModel, amount: BigNumber): OutputModel =>
+      new OutputModel({
+        asset: output.asset,
+        value: output.value.add(clientUtils.bigNumberToBN(amount, 8)),
+        address: output.address,
+      });
+
+    const consoladatableUnspentsIn = await this.provider.getUnspentOutputs(from.network, from.address, monitor);
+
+    const looseAssets = consoladatableUnspentsIn
```
dicarlo2: I would call this `assetToUnspentOutputs`. I have no idea what `looseAssets` means, but `assetToUnspentOutputs` means exactly what it says, it's a map from asset to unspent outputs.

packages/neo-one-client/src/user/LocalUserAccountProvider.ts
```diff
@@ -823,6 +841,172 @@ export class LocalUserAccountProvider<TKeyStore extends KeyStore, TProvider exte
...
+      throw new TransactionSizeExceedsLimitError(messageSize, byteLimit);
+    }
+    if (byteLimit - messageSize - newInputSize < 0) {
+      return transactionUnsignedIn;
+    }
+
+    const inputMap = (): { [key: string]: ReadonlyArray<UnspentOutput> } => ({});
+    const inputArray = (): ReadonlyArray<InputModel> => [];
+    const outputArray = (): ReadonlyArray<OutputModel> => [];
+    const getMessageSize = (numInputs: number, numOutputs: number) =>
```
dicarlo2: It's generally a good idea to use an object as an input when you have multiple primitive arguments. As it stands, there's nothing to stop you from mixing up `numInputs` and `numOutputs`, whereas if the input was `{ numInputs: number; numOutputs: number }`, it'd be much more difficult to make a mistake. Also much clearer for the reader that you're passing the right arguments.

dicarlo2: Also, this should probably be `numNewInputs` and `numNewOutputs` - it's not the total number of inputs/outputs of the transaction.

packages/neo-one-client/src/user/LocalUserAccountProvider.ts
```diff
@@ -823,6 +841,172 @@ export class LocalUserAccountProvider<TKeyStore extends KeyStore, TProvider exte
...
+      .filter(
+        (output) =>
+          (looseAssets[common.uInt256ToString(output.asset)] as ReadonlyArray<UnspentOutput> | undefined) !==
+            undefined && common.uInt160ToString(output.address) === addressToScriptHash(from.address),
+      )
+      .reduce(
+        (acc, output) => {
+          const currentAsset = common.uInt256ToString(output.asset);
+          const toAdd = Math.min(
+            looseAssets[currentAsset].length,
```
dicarlo2: This should be `acc.looserAssets`

packages/neo-one-client/src/user/LocalUserAccountProvider.ts
```diff
@@ -823,6 +841,172 @@ export class LocalUserAccountProvider<TKeyStore extends KeyStore, TProvider exte
...
+            Math.floor((byteLimit - getMessageSize(acc.newInputs.length, 0)) / newInputSize),
+          );
+          if (toAdd === 0) {
```
dicarlo2: This early return isn't necessary, remove it.

packages/neo-one-client/src/user/LocalUserAccountProvider.ts
```diff
@@ -823,6 +841,172 @@ export class LocalUserAccountProvider<TKeyStore extends KeyStore, TProvider exte
...
+            return acc;
+          }
+          const tempIns = acc.looserAssets[currentAsset]
+            .sort((coinA, coinB) => coinA.value.comparedTo(coinB.value))
```
dicarlo2: IIRC this sorts in ascending order, is that the order we want? There should be a test case for this.

packages/neo-one-client/src/user/LocalUserAccountProvider.ts
```diff
@@ -823,6 +841,172 @@ export class LocalUserAccountProvider<TKeyStore extends KeyStore, TProvider exte
...
+            .slice(0, toAdd);
+          const newUpdatedOutputs = acc.updatedOutputs.map((out) => {
```
dicarlo2: This is odd, why are we iterating over the outputs again? Definitely a code smell to see effectively two loops, there should be a better way to structure this. Why not just do something like
```
return {
  updatedOutputs: acc.updatedOutputs.concat([updateOutput(output, ...)]),
}
```

You already have the output you want to change, so just change it. Then change the `updatedOutputs` initializer to `updatedOutputs: []` instead. Also remove the initial filter and just do it inside the `reduce` function.

packages/neo-one-client/src/user/LocalUserAccountProvider.ts
dicarlo2: Probably should have a test that catches this bug.

packages/neo-one-client/src/user/LocalUserAccountProvider.ts
dicarlo2: Also sorting on every iteration seems odd too. `O(n^2logn)` complexity. It doesn't really matter in this case, but generally you should be trying to write optimal code so that it's structured consistently. Instead, just sort above (see comment).

packages/neo-one-client/src/user/LocalUserAccountProvider.ts
```diff
@@ -823,6 +841,172 @@ export class LocalUserAccountProvider<TKeyStore extends KeyStore, TProvider exte
...
+      });
+
+    const consoladatableUnspentsIn = await this.provider.getUnspentOutputs(from.network, from.address, monitor);
+
+    const looseAssets = consoladatableUnspentsIn
+      .filter(
+        (unspent) =>
+          !inputs.some((input) => unspent.txid === input.txid && unspent.vout === input.vout) &&
+          !this.mutableUsedOutputs.has(`${unspent.txid}:${unspent.vout}`),
+      )
```
dicarlo2: `.sort((outputA, outputB) => outputA.value.comparedTo(outputB.value))`

packages/neo-one-client/src/user/LocalUserAccountProvider.ts
dicarlo2: Also, this variable seems unnecessary.

packages/neo-one-client/src/user/LocalUserAccountProvider.ts
```diff
@@ -823,6 +841,172 @@ export class LocalUserAccountProvider<TKeyStore extends KeyStore, TProvider exte
...
+          };
+
+          return {
+            newInputs: acc.newInputs.concat(this.convertInputs(tempIns)),
+            updatedOutputs: newUpdatedOutputs,
+            looserAssets: evenLooserAssets,
+          };
+        },
+        { newInputs: inputArray(), updatedOutputs: transactionUnsignedIn.outputs, looserAssets: looseAssets },
+      );
```
dicarlo2: Lots of comments above, roughly, here's how I'd do it:
```
const { newInputs, newOutputs, remainingAssetToUnspentOutputs } = transactionUnsignedIn.outputs.reduce(
  (acc, output) => {
    const asset = common.uInt256ToString(output.asset);
    const unspentOutputs = (
      (acc.remainingAssetToUnspentOutputs[asset] as ReadonlyArray<UnspentOutput> | undefined) === undefined ||
      common.uInt160ToString(output.address) !== addressToScriptHash(from.address)
    ) 
      ? []
      : acc.remainingAssetToUnspentOutputs[asset;
    const toAdd = Math.floor((byteLimit - getMessageSize({ numNewInputs: acc.newInputs.length })) / InputModel.size);
    const newInputs = unspentOutputs.slice(0, toAdd);

    return { 
      newInputs: acc.newInputs.concat(this.convertInputs(newInputs)),
      newOutputs: acc.newOutputs.concat([
        output.clone({ 
          value: output.value.plus(newInputs.reduce((acc, input) => acc.plus(input.value), new BigNumber('0'))),
        }),
      ]),
      remainingAssetToUnspentOutputs: {
        ...acc.remainingAssetToUnspentOutputs,
        [asset]: unspentOutputs.slice(toAdd),
      },
    };
  }, 
  { newInputs: [], newOutputs: [], remainingAssetToUnspentOutputs: assetToUnspentOutputs },
);
```

dicarlo2: Might need to add an explicit type to `reduce` so that typescript doesn't complain. (something like `reduce<{ newInputs: ReadonlyArray<Input>, ...}>`)

packages/neo-one-client/src/user/LocalUserAccountProvider.ts
```diff
@@ -823,6 +841,172 @@ export class LocalUserAccountProvider<TKeyStore extends KeyStore, TProvider exte
...
+
+    const { newestInputs, newOutputs } = Object.keys(looserAssets)
+      .filter((key) => looserAssets[key].length >= 2)
+      .reduce(
+        (acc, key) => {
+          if (byteLimit - getMessageSize(acc.newestInputs.length + 2, acc.newOutputs.length + 1) < 0) {
```
dicarlo2: Same here, remove this.

packages/neo-one-client/src/user/LocalUserAccountProvider.ts
```diff
@@ -823,6 +841,172 @@ export class LocalUserAccountProvider<TKeyStore extends KeyStore, TProvider exte
...
+              ),
+            );
+          const newOutput: Output = {
+            asset: key,
+            value: newUnspents.reduce((left, right) => left.plus(right.value), new BigNumber('0')),
+            address: from.address,
+          };
+
+          return {
+            newestInputs: acc.newestInputs.concat(this.convertInputs(newUnspents)),
```
dicarlo2: `newInputs`

packages/neo-one-client/src/user/LocalUserAccountProvider.ts
```diff
@@ -823,6 +841,172 @@ export class LocalUserAccountProvider<TKeyStore extends KeyStore, TProvider exte
...
+          return {
+            newInputs: acc.newInputs.concat(this.convertInputs(tempIns)),
+            updatedOutputs: newUpdatedOutputs,
+            looserAssets: evenLooserAssets,
+          };
+        },
+        { newInputs: inputArray(), updatedOutputs: transactionUnsignedIn.outputs, looserAssets: looseAssets },
+      );
+
+    const { newestInputs, newOutputs } = Object.keys(looserAssets)
```
dicarlo2: Do something like `const { newInputs: finalInputs, newOutputs: finalOutputs } = `

dicarlo2: `Object.entries(assetToUnspentOutput)`

packages/neo-one-client/src/user/LocalUserAccountProvider.ts
```diff
@@ -823,6 +841,172 @@ export class LocalUserAccountProvider<TKeyStore extends KeyStore, TProvider exte
...
+      .reduce(
+        (acc, key) => {
+          if (byteLimit - getMessageSize(acc.newestInputs.length + 2, acc.newOutputs.length + 1) < 0) {
+            return acc;
+          }
+          const newUnspents = looserAssets[key]
+            .sort((coinA, coinB) => coinA.value.comparedTo(coinB.value))
+            .slice(
+              0,
+              Math.min(
```
dicarlo2: Remove, you just need `Math.floor`

packages/neo-one-client/src/user/LocalUserAccountProvider.ts
```diff
@@ -823,6 +841,172 @@ export class LocalUserAccountProvider<TKeyStore extends KeyStore, TProvider exte
...
+
+    const { newestInputs, newOutputs } = Object.keys(looserAssets)
+      .filter((key) => looserAssets[key].length >= 2)
+      .reduce(
+        (acc, key) => {
+          if (byteLimit - getMessageSize(acc.newestInputs.length + 2, acc.newOutputs.length + 1) < 0) {
+            return acc;
+          }
+          const newUnspents = looserAssets[key]
+            .sort((coinA, coinB) => coinA.value.comparedTo(coinB.value))
```
dicarlo2: Remove.

packages/neo-one-client/src/user/LocalUserAccountProvider.ts
```diff
@@ -823,6 +841,172 @@ export class LocalUserAccountProvider<TKeyStore extends KeyStore, TProvider exte
...
+            .slice(
+              0,
+              Math.min(
+                looseAssets[key].length,
+                Math.floor(
```
dicarlo2: I'd pull this out like you did above.

packages/neo-one-client/src/user/LocalUserAccountProvider.ts
```diff
@@ -823,6 +841,172 @@ export class LocalUserAccountProvider<TKeyStore extends KeyStore, TProvider exte
...
+      );
+
+    const { newestInputs, newOutputs } = Object.keys(looserAssets)
+      .filter((key) => looserAssets[key].length >= 2)
+      .reduce(
+        (acc, key) => {
+          if (byteLimit - getMessageSize(acc.newestInputs.length + 2, acc.newOutputs.length + 1) < 0) {
+            return acc;
+          }
+          const newUnspents = looserAssets[key]
```
dicarlo2: `const newOutputs = outputs.slice(...)`

packages/neo-one-client/src/user/LocalUserAccountProvider.ts
```diff
@@ -823,6 +841,172 @@ export class LocalUserAccountProvider<TKeyStore extends KeyStore, TProvider exte
...
+                looseAssets[key].length,
+                Math.floor(
+                  (byteLimit - getMessageSize(acc.newestInputs.length, acc.newOutputs.length) - newOutputSize) /
+                    newInputSize,
+                ),
+              ),
+            );
+          const newOutput: Output = {
+            asset: key,
+            value: newUnspents.reduce((left, right) => left.plus(right.value), new BigNumber('0')),
```
dicarlo2: `.reduce((acc, output) => acc.plus(output.value), ...),`

packages/neo-one-client/src/user/LocalUserAccountProvider.ts
```diff
@@ -823,6 +841,172 @@ export class LocalUserAccountProvider<TKeyStore extends KeyStore, TProvider exte
...
+            newInputs: acc.newInputs.concat(this.convertInputs(tempIns)),
+            updatedOutputs: newUpdatedOutputs,
+            looserAssets: evenLooserAssets,
+          };
+        },
+        { newInputs: inputArray(), updatedOutputs: transactionUnsignedIn.outputs, looserAssets: looseAssets },
+      );
+
+    const { newestInputs, newOutputs } = Object.keys(looserAssets)
+      .filter((key) => looserAssets[key].length >= 2)
```
dicarlo2: `.filter(([_asset, outputs]) => outputs.length >= 2)`

packages/neo-one-client/src/user/LocalUserAccountProvider.ts
```diff
@@ -823,6 +841,172 @@ export class LocalUserAccountProvider<TKeyStore extends KeyStore, TProvider exte
...
+      .reduce(
+        (acc, key) => {
```
dicarlo2: `(acc, [asset, outputs]) => {`

packages/neo-one-client/src/user/LocalUserAccountProvider.ts
```diff
@@ -823,6 +841,172 @@ export class LocalUserAccountProvider<TKeyStore extends KeyStore, TProvider exte
...
+
+          return {
+            newestInputs: acc.newestInputs.concat(this.convertInputs(newUnspents)),
+            newOutputs: acc.newOutputs.concat(this.convertOutputs([newOutput])),
+          };
+        },
+        { newestInputs: newInputs, newOutputs: outputArray() },
+      );
+
+    switch (transactionUnsignedIn.type) {
```
dicarlo2: This should just call `transactionUnsigned.clone()` and let the subclasses deal with it.

packages/neo-one-client/src/user/LocalUserAccountProvider.ts
```diff
@@ -823,6 +841,172 @@ export class LocalUserAccountProvider<TKeyStore extends KeyStore, TProvider exte
...
+      case TransactionType.Issue:
+        return new IssueTransaction({
+          version: transactionUnsignedIn.version,
+          attributes: transactionUnsignedIn.attributes,
+          inputs: transactionUnsignedIn.inputs.concat(newestInputs),
+          outputs: updatedOutputs.concat(newOutputs),
+          scripts: transactionUnsignedIn.scripts,
+          hash: transactionUnsignedIn.hash,
+        });
+      default:
```
dicarlo2: Why only these transaction types?

packages/neo-one-client/src/user/LocalUserAccountProvider.ts
```diff
@@ -823,6 +841,172 @@ export class LocalUserAccountProvider<TKeyStore extends KeyStore, TProvider exte
...
   }
 
+  private async consolidate({
+    inputs,
+    transactionUnsignedIn,
+    from,
+    monitor,
+    byteLimit,
+  }: {
+    readonly transactionUnsignedIn: TransactionModel;
```
dicarlo2: `transactionUnsigned`

packages/neo-one-client/src/user/LocalUserAccountProvider.ts
danwbyrne: I actually can't, when I save it formats it like this. Originally I had tried to do it with parentheses.

packages/neo-one-client-core/src/transaction/Input.ts
```diff
@@ -22,6 +22,7 @@ export interface InputJSON {
 }
 
 export class Input implements SerializableWire<Input>, Equatable, SerializableJSON<InputJSON> {
+  public static readonly size: number = IOHelper.sizeOfUInt256 + IOHelper.sizeOfUInt16LE;
```
dicarlo2: Make this a `const` at module scope similar to `Output`.

packages/neo-one-client/src/__tests__/user/LocalUserAccountProvider.test.ts
```diff
@@ -1076,4 +1097,210 @@ describe('LocalUserAccountProvider', () => {
...
+    });
+    const consolidatedTransaction = spy.mock.calls[0][0].transaction;
+    expect(consolidatedTransaction.serializeUnsigned().length).toEqual(366);
+    expect(consolidatedTransaction.inputs.length).toEqual(5);
+    expect(consolidatedTransaction.outputs.length).toEqual(3);
+
+    spy.mockReset();
+    spy.mockRestore();
+    verifyMocks();
+  });
```
dicarlo2: Add a newline

packages/neo-one-client/src/__tests__/user/LocalUserAccountProvider.test.ts
```diff
@@ -1076,4 +1097,210 @@ describe('LocalUserAccountProvider', () => {
...
+    ];
+    keystore.byteLimit = 300;
+    keystore.sign = jest.fn(() => witness);
+    provider.getUnspentOutputs = jest.fn(() => Promise.resolve(unspent));
+    provider.getTransactionReceipt = jest.fn(() => Promise.resolve('receiptDummy'));
+
+    const result = localUserAccountProvider.transfer(transfer, {
+      from: options.from,
+      attributes: options.attributes,
+      networkFee: undefined,
```
dicarlo2: Remove `networkFee: undefined` everywhere

packages/neo-one-client/src/user/LocalUserAccountProvider.ts
```diff
@@ -823,6 +839,172 @@ export class LocalUserAccountProvider<TKeyStore extends KeyStore, TProvider exte
...
+        const potentialBytes =
+          byteLimit -
+          getMessageSize({ numNewInputs: acc.finalInputs.length + 2, numNewOutputs: acc.newOutputs.length + 1 });
+
+        if (potentialBytes < 0) {
+          return acc;
+        }
+        const newUnspents = outputs.slice(
+          0,
+          Math.floor(
```
dicarlo2: `Math.max`

packages/neo-one-client/src/user/LocalUserAccountProvider.ts
```diff
@@ -823,6 +839,172 @@ export class LocalUserAccountProvider<TKeyStore extends KeyStore, TProvider exte
...
+      },
+    ).reduce(
+      (
+        acc: {
+          readonly finalInputs: ReadonlyArray<InputModel>;
+          readonly newOutputs: ReadonlyArray<OutputModel>;
+        },
+        [asset, outputs],
+      ) => {
+        const potentialBytes =
```
dicarlo2: Remove these and put a length check below on `newUnspents`.

packages/neo-one-client/src/user/LocalUserAccountProvider.ts
```diff
@@ -823,6 +839,172 @@ export class LocalUserAccountProvider<TKeyStore extends KeyStore, TProvider exte
...
+  }: {
+    readonly transactionUnsignedIn: TransactionModel;
+    readonly inputs: ReadonlyArray<UnspentOutput>;
+    readonly from: UserAccountID;
+    readonly monitor?: Monitor;
+    readonly byteLimit: number;
+  }): Promise<TransactionModel> {
+    const messageSize = transactionUnsignedIn.serializeUnsigned().length;
+
+    if (byteLimit - messageSize < 0) {
```
dicarlo2: Probably don't need these conditions

Issue #313, opened 2018-07-10
Contributor recognition


dicarlo2:
https://github.com/kentcdodds/all-contributors

PR #325, opened 2018-07-15
Contributor recognition

speenzor:
### Requirements

### Description of the Change

#313 

Adds grid of contributors in the README and sets up the project for further use of the all-contributors-cli package on other machines. The all-contributors-cli package through npm is an easy way to add contributors in the future with only a few easy commands.

All-contributors: https://github.com/kentcdodds/all-contributors
All-contributors-cli package: https://www.npmjs.com/package/all-contributors-cli#installation

The CLI package adds `.all-contributorsrc` and `package-lock.json` in the root directory as well as adding two commands in "scrips" in `package.json`. `all-contributorsrc` keeps track of contributors, their photos, contributions, and other meta data. The commands are for adding contributors and for generating the contributor table.

### Test Plan

No testing is necessary as far as I can tell. The new `package-lock.json` file is the only area where I'm uncertain as to its affects.

### Alternate Designs

The all-contributors table can also be done in a separate file from the README and can also be created by hand as opposed to created using the CLI package. This would make the all-contributors table harder to maintain down the line as the number of contributors grows. The all-contributors project also has an Atom package, but I did not look closely at that as the CLI npm package was elegant enough and easy to use.

### Benefits

Recognizing contributors.

### Possible Drawbacks

The CLI package created the `package-lock.json` file. I'm still not entirely sure what this is for, if it's necessary, or if it's even desirable to have. I'm open to feedback on this if it's a problem.

### Applicable Issues

None that I can tell.


package.json
```diff
@@ -118,6 +120,7 @@
     "@types/tar": "4.0.0",
     "@types/through": "0.0.29",
     "@types/yargs": "11.1.0",
+    "all-contributors-cli": "^5.2.1",
```
DalderupMaurice: Seems like the only dep which isn't pinned ^^

speenzor: You're right. Thank you for pointing that out. Changing that now.

README.md
```diff
@@ -10,3 +11,15 @@ Welcome to the NEO•ONE community! We're always looking for more contributors a
 ### License
```
dicarlo2: Move this to the end and make it a `##` heading.

PR #340, opened 2018-07-20
fixes build errors

danwbyrne:
removes `import * as crypto from 'crypto'` from some places of client-core which caused a deprecation warning to trigger even though we never used the deprecated modules.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,6 +8,7 @@
 /packages/*/dist
 /packages/*/dist-ts
 .rpt2_cache
+/reports
 
 # Logs
 logs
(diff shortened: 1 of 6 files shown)


package.json
```diff
@@ -53,7 +53,7 @@
     "test": "jest --config jest/unit.js",
     "test-ci": "jest --config jest/unit-ci.js --ci -w 1",
     "e2e": "jest --config jest/e2e.js",
-    "e2e-ci": "jest --config jest/e2e-ci.js --ci -w 1",
+    "e2e-ci": "node --no-warnings ./node_modules/.bin/jest --config jest/e2e-ci.js --ci -w 1",
```
dicarlo2: We should leave warnings on and instead eliminate the warnings everywhere.

danwbyrne: You can ignore the no-warnings I was just seeing if it changed something with CI.

All `import * as <> from ‘crypto’` are out of the code base.

PR #349, opened 2018-07-27
Adds code coverage reports to codecov.io and circleci

danwbyrne:
### Description of the Change

We now use `nyc` and to generate code coverage reports. This is initial testing but it seems pretty plug and play. 

### Test Plan

```bash
#collects coverage results while testing
yarn nyc --clean <any jest test> --coverage
#logs out the most recent coverage report
yarn nyc report
#sends the coverage results to codecov.io
yarn posttest --token=<codecoverage token>
```
running these you should see code coverage reports for the given test.

I've added `nyc` to `package.json` as well as `codecov`. 

Then in CI we do:
```
- run:
      name: Test
      command: yarn nyc --clean yarn test-ci --coverage
- run:
      name: Generate Code Coverage - UNIT
      command: yarn posttest --token=<codecoverage token> // <-- using mine for now
- run:
      name: E2E
      command: yarn nyc --clean yarn e2e-ci --coverage
- run:
      name: Generate Code Coverage - E2E
      command: yarn posttest --token=<codecoverage token>
```

### Benefits

Code coverage is SICK.

### Possible Drawbacks

I am seeing a memory issue on initial tests. We'll figure out whats going on in due time.

### Applicable Issues

I think https://github.com/istanbuljs/nyc/issues/729#issuecomment-370781801 fixes all of our issues with jest coverage generating with `nyc`.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -53,7 +53,7 @@ aliases:
       - save-cache: *save-yarn-cache
       - run:
           name: Test
-          command: yarn test-ci $(circleci tests glob "packages/*/src/__tests__/**/*.test.ts" | circleci tests split --split-by=timings)
+          command: yarn nyc --clean test-ci $(circleci tests glob "packages/*/src/__tests__/**/*.test.ts" | circleci tests split --split-by=timings) --coverage
       - store_test_results:
           path: reports
     parallelism: 2
@@ -82,7 +82,7 @@ aliases:
           at: ./
       - run:
           name: E2E
-          command: yarn e2e-ci $(circleci tests glob "packages/*/src/__e2e__/**/*.test.ts" | circleci tests split --split-by=timings)
+          command: yarn nyc --clean e2e-ci $(circleci tests glob "packages/*/src/__e2e__/**/*.test.ts" | circleci tests split --split-by=timings) --coverage
       - store_test_results:
           path: reports
     parallelism: 2
--- a/package.json
+++ b/package.json
@@ -55,6 +55,9 @@
     "e2e": "jest --config jest/e2e.js",
     "e2e-ci": "jest --config jest/e2e-ci.js --ci -w 1",
     "tsc": "tsc",
+    "nyc": "./node_modules/.bin/nyc",
+    "codecov": "./node_modules/.bin/codecov",
+    "posttest": "yarn run nyc report --report=json && yarn run codecov -f coverage/*.json",
     "ts-node": "cross-env NODE_OPTIONS=\"--max-old-space-size=8192\" TS_NODE_PROJECT=tsconfig/tsconfig.es2018.cjs.json node -r ts-node/register/transpile-only",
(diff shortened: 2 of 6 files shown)


package.json
```diff
@@ -55,6 +55,9 @@
     "e2e": "jest --config jest/e2e.js",
     "e2e-ci": "jest --config jest/e2e-ci.js --ci -w 1",
     "tsc": "tsc",
+    "nyc": "./node_modules/.bin/nyc",
```
dicarlo2: Remove, this isn't necessary - this already happens automatically.

package.json
```diff
@@ -55,6 +55,9 @@
...
+    "codecov": "./node_modules/.bin/codecov",
```
dicarlo2: Remove, this isn't necessary - this already happens automatically.

package.json
```diff
@@ -55,6 +55,9 @@
...
+    "posttest": "yarn run nyc report --report=json && yarn run codecov -f coverage/*.json",
```
dicarlo2: Don't call it `posttest` - `post` and `pre` have special meaning in a package.json - basically when you run `yarn foo`, first it checks to see if `prefoo` exists - if it does, it runs it. Then it runs `foo`, and finally `postfoo` if it exists.

package.json
```diff
@@ -81,6 +84,13 @@
       "post-rewrite": "yarn run ts-node ./scripts/hooks/post-rewrite.ts"
     }
   },
+  "jest": {
```
dicarlo2: Remove, you should add additional jest configuration in the `jest` subfolder. You want to make additional configurations like `test-ci-coverage.js` and `e2e-ci-coverage.js` that are a combination of `test-ci` and `e2e-ci` (respectively) + coverage configuration.

package.json
```diff
@@ -81,6 +84,13 @@
...
+    "coverageReporters": ["json"]
+  },
+  "nyc": {
```
dicarlo2: Move to its own configuration file (assuming nyc supports configuration files vs package.json configuration)

tsconfig.json
```diff
@@ -41,6 +41,7 @@
     "alwaysStrict": true,
     "strict": true,
     "skipLibCheck": true,
+    "sourceMap": true,
```
dicarlo2: Remove. Point nyc at the appropriate configuration in the `tsconfig` subfolder. Alternatively or possibly in addition, you should pre-instrument the code by adding a build configuration that outputs the code required for e2e testing so that it runs faster.

.circleci/config.yml
```diff
@@ -45,10 +45,17 @@ aliases:
           command: yarn lint
       - run:
           name: Test
-          command: yarn test-ci
+          command: yarn nyc --clean yarn test-ci --coverage
```
dicarlo2: Actually, I don't think we need to run nyc for unit tests since they don't (or shouldn't) launch any sub processes.

Issue #397, opened 2018-08-14
Fix Node 8/9 CI Build


danwbyrne:
## Package.json
https://stackoverflow.com/questions/43694281/ts2318-cannot-find-global-type-asynciterableiterator-async-generator
is the fix
basically we need to run with the flag --harmony_async_iteration if the node version is 8
and --harmony if the node version is 9

## GULP
just need to add the appropriate harmony environment flag
there should be an equivalent for each version for the cli flags
that one should be easier because you can just check one of the environment variables for the node version
like process.env.NODE_VERSION or something

PR #398, opened 2018-08-14
fixes async iteration failure on older node versions

danwbyrne:
### Requirements

NEW DEPENDENCIES OH BOY:
```
"@reactivex/ix-esnext-esm":"2.3.5",
```
for the build.

### Description of the Change

just a bug fix that adjusts some scripts to allow asyncIterators to be used even when on older node versions (8, 9).

### Test Plan

IT BUILDS AND PASSES TESTS

### Alternate Designs

might make `asyncFlag.sh` into a node script, similar to gulp (but way simpler).

### Benefits

G R E E N

### Possible Drawbacks

Our red constituents will not be pleased.

### Applicable Issues

#397


scripts/asyncFlag.js
```diff
@@ -0,0 +1,16 @@
+const execa = require('execa');
+
+const cmd = async () =>
```
dicarlo2: You can use `process.version`, no need to shell out.

Issue #421, opened 2018-08-23
Convert `CustomError` to `makeErrorWithCode`


danwbyrne:
I'm not 100% on the reasoning but I do as I'm told. We get rid of the old `CustomError` class implementation of our errors and replace it with the `makeErrorWithCode` function.

Issue #420, opened 2018-08-23
Overhaul Errors


danwbyrne:
I'm going to commit the converted errors now with a fixed build, then I will overhaul all of them (immediately, should be done today). I mainly just want the converted Errors in now because it completely removes `CustomError`.

PR #423, opened 2018-08-24
Errors

danwbyrne:
### Description of the Change

Handled 2 issues here:
1) convert all of our old errors of the form `class <error> extends CustomError` to `const <error> = makeErrorWithCode(...)`.
2) overhaul error messages to be more informative. 

### Test Plan

Tests run the same, although since I changed the error messages on a few things I had to update some snapshots.

### Alternate Designs

Potentially with this `makeErrorWithCode` function we could have an `extraInfo` optional property that we could hide/reveal longer error messages if we wanted to. Just an idea.

### Benefits

errors are better now.

### Applicable Issues

#421 
#420


packages/neo-one-client-core/src/ContractPropertyState.ts
```diff
@@ -35,17 +35,10 @@ export const HasPayable = new Set([
...
-
-  public constructor(contractParameterType: number) {
-    super(`Expected contract parameter type, ` + `found: ${contractParameterType.toString(16)}`);
-
-    this.contractParameterType = contractParameterType;
-    this.code = 'INVALID_ContractPropertyState';
-  }
-}
+export const InvalidContractPropertyStateError = makeErrorWithCode(
+  'INVALID_ContractPropertyState',
```
dicarlo2: oops, fix this to be in all caps `INVALID_CONTRACT_PROPERTY_STATE`

packages/neo-one-client-core/src/crypto/crypto.ts
```diff
@@ -166,14 +152,10 @@ const verify = ({
...
-  public readonly code: string;
-
-  public constructor() {
-    super('Invalid Private Key');
-    this.code = 'INVALID_PRIVATE_KEY';
-  }
-}
+export const InvalidPrivateKeyError = makeErrorWithCode(
+  'INVALID_PRIVATE_KEY',
+  (privateKey: PrivateKey) => `Invalid Private Key, found: ${privateKey}`,
```
dicarlo2: Convert to a string

Issue #415, opened 2018-08-22
Ledger 2.0


danwbyrne:
While we are at a solid base for ledger support, there is a few more things I would like to do in the future:
1) hook up the node-transport's `listen` function so we can plug-and-play with ledgers.
2) emulate a `listen` function for u2f-transport.
3) move scanning/updating into the ledger handler itself. I haven't really fleshed out this idea but I think it would be nice to scan on the fly instead of all at once when we `.init()`.
4) we could cache `account#` -> `publicKey` on each ledger if we are going to be scanning on the fly.
5) allow configurable names.

these are just ideas I've had, none are super fleshed out atm.

dicarlo2:
@danwbyrne can you split this issue out into separate cards? Going to archive for now.

PR #403, opened 2018-08-16
adds a new LedgerKeyStore for U2F and Node Ledger access

danwbyrne:
# 【﻿Ｌｅｄｇｅｒ　Ｓｕｐｐｏｒｔ】
#192 
## Description of the Change

Strap in for this commit boys. L E D G E R support is here! (almost)

In this commit I've done a few things:
1) created declaration files for external ledger packages in our `@types` directory.
2) created a browser/node switchable `LedgerTransport` in `@neo-one/client-switch` which handles communicating with our Ledgers in their respective environments.
3) created `LedgerKeyStore` in `@neo-one/client/user/keystores/` which implements the KeyStore interface (modified to fit some functions that aren't necessary for ledger) using our newly defined `LedgerHandler` to handle the passing of messages and connections to ledgers.
4) ??? hopefully this just hooks up super ez-pz?
5) profit.

## Requirements

we add two new dependencies to `@neo-one/client-switch`,
```json
    "@ledgerhq/hw-transport-node-hid":"4.22.0",
    "@ledgerhq/hw-transport-u2f":"4.21.0",
```

additionally I have created declaration files for the Transport classes inside these at `@types/@ledgerhq/`

### KeyStores
Need to test `LedgerKeyStore` in the same manner that we test `LocalKeyStore`.

*Update* I have added tests for `LedgerKeyStore`, need to get with you @dicarlo2 and review all of this. I think there is still a few more spots for improvement but I'd like to touch base with you again. It is working for our ledger though. :)

### Transports
I think the one thing I want to make certain of now for Transports is that we are handling the opening/closing of them properly. Also I think we could definitely hook up the `listen` observable from Transport now to easily support more ledgers in our `LedgerKeyStore`. Something to discuss.

## Alternate Designs

It spent a lot of time figuring out the best way to abstract Keystores / Transports / Stores / AccountProviders and while it seemed like a good idea originally, ultimately we went with just creating a new `LedgerKeyStore` class alongside `LocalKeyStore` that has similar functionality but doesn't use the concept of *wallets*.

## Benefits

Ｌｅｄｇｅｒ　Ｓｕｐｐｏｒｔ

## Possible Drawbacks

To all of our fans out their who were looking for a wallet that **did not** support ledgers, this is going to be a major turn-off.

## Applicable Issues

**BIG ISSUE** can't get it to work on CI, something wrong with a dependency set up by the docker file.

TODO: custom errors

Issue #192, opened 2018-06-11
Add ledger support


dicarlo2:

dicarlo2:
We can add a new [`KeyStore`](https://github.com/neo-one-suite/neo-one/blob/master/packages/neo-one-client/src/user/LocalUserAccountProvider.ts#L80) implementation that supports ledger, say `LedgerKeyStore`. See [`LocalKeyStore`](https://github.com/neo-one-suite/neo-one/blob/master/packages/neo-one-client/src/user/keystore/LocalKeyStore.ts) for an example of an implementation of the interface. Effectively what it should do is extract the available accounts (addresses) from the ledger device to populate the various account methods as well as locally store things like the account name and which accounts have been "deleted". It should implement `sign` using the Ledger api as well.

We'll want to target both the Ledger api for node and for the browser and the best way to do this in neo-one is to create a separate package with entrypoints for both node and the browser. The structure should look something like this:

- packages
    - neo-one-client-ledger-keystore
        - src
            - index.ts
            - index.browser.ts
            - LedgerNodeKeyStore.ts
            - LedgerBrowserKeyStore.ts
        - package.json
        - tsconfig.json
 
Where index.ts looks like
```typescript
export { LedgerNodeKeyStore as LedgerKeyStore } from './LedgerNodeKeyStore';
```

and index.browser.ts looks like

```typescript
export { LedgerBrowserKeyStore as LedgerKeyStore } from './LedgerBrowserKeyStore';
```

The ledger browser api can be seen in use [here](https://github.com/leeyikjiun/blue-app-neo-tests/blob/master/test/u2f.js). The ledger node api [here](https://github.com/leeyikjiun/blue-app-neo-tests/blob/master/test/node-hid.js). They're effectively the same, so you'll likely want to make a base class `LedgerKeyStore` that takes in which one of the apis to use. Then your `LedgerNodeKeyStore` and `LedgerBrowserKeyStore` just subclass `LedgerKeyStore` and call `super` with the appropriate api.

Use the Ledger from my desk to test. You should write unit tests just like normal, but put them under a different directory like `__tests_ledger__` so that they are not automatically run with all of the other tests. Then create a `test-ledger` script in the root `package.json` similar to the `e2e` script that runs your unit tests so that we can trigger them manually.

Also, be sure to write normal unit tests under `__test__` that test your base class with a mock ledger interface.

danwbyrne:
https://github.com/LedgerHQ/ledgerjs

[...]

PR #490, opened 2018-09-27
Ledger Support for Node and Browser

danwbyrne:
# 【﻿Ｌｅｄｇｅｒ　Ｓｕｐｐｏｒｔ】
#192 

(resubmission of #403 after refactoring of neo-one)
## Description of the Change

Strap in for this commit boys. L E D G E R support is here!

In this commit I've done a few things:
1) created declaration files for external ledger packages in our `@types` directory.
2) created a browser/node switchable `LedgerTransport` in `@neo-one/client-switch` which handles communicating with our Ledgers in their respective environments.
3) created `LedgerKeyStore` in `@neo-one/client/user/keystores/` which implements the KeyStore interface (modified to fit some functions that aren't necessary for ledger) using our newly defined `LedgerHandler` to handle the passing of messages and connections to ledgers.
4) ??? hopefully this just hooks up super ez-pz?
5) profit.

## Requirements

we add two new dependencies to `@neo-one/client-switch`,
```json
    "@ledgerhq/hw-transport-node-hid":"4.22.0",
    "@ledgerhq/hw-transport-u2f":"4.21.0",
```

additionally I have created declaration files for the Transport classes inside these at `@types/@ledgerhq/`

### KeyStores
Need to test `LedgerKeyStore` in the same manner that we test `LocalKeyStore`.

*Update* I have added tests for `LedgerKeyStore`, need to get with you @dicarlo2 and review all of this. I think there is still a few more spots for improvement but I'd like to touch base with you again. It is working for our ledger though. :)

### Transports
I think the one thing I want to make certain of now for Transports is that we are handling the opening/closing of them properly. Also I think we could definitely hook up the `listen` observable from Transport now to easily support more adding/removing ledger-path connections in our `LedgerKeyStore`. Something to discuss.

## Alternate Designs

It spent a lot of time figuring out the best way to abstract Keystores / Transports / Stores / AccountProviders and while it seemed like a good idea originally, ultimately we went with just creating a new `LedgerKeyStore` class alongside `LocalKeyStore` that has similar functionality but doesn't use the concept of *wallets*.

## Benefits

Ｌｅｄｇｅｒ　Ｓｕｐｐｏｒｔ

## Possible Drawbacks

To all of our fans out their who were looking for a wallet that **did not** support ledgers, this is going to be a major turn-off.

## Applicable Issues

Getting it working on CI has been a struggle.

TODO: #415

### DIFF
--- a/packages/neo-one-client-core/src/user/keystore/LedgerHandler.ts
+++ b/packages/neo-one-client-core/src/user/keystore/LedgerHandler.ts
@@ -1,0 +1,134 @@
+import { crypto, PublicKeyString } from '@neo-one/client-common';
+import { HWLedger, LedgerTransport } from '@neo-one/client-switch';
+import { LedgerNotDetectedError, LedgerNotSupportedError, LedgerStatusCodeError } from '../../errors';
+
+export interface LedgerQueueRequest {
+  readonly msg: Buffer;
+  readonly resolve: (value: Buffer) => void;
+  readonly reject: (error: Error) => void;
+}
+
+const keyRequestParams = Buffer.from('80040000', 'hex');
+const initMessageParams = Buffer.from('80020000', 'hex');
+const finalMessageParams = Buffer.from('80028000', 'hex');
+
+const BIP44 = (accountIn = 0): Buffer => {
+  const account = accountIn.toString(16);
+
+  return Buffer.from(`8000002C800003788000000000000000${'0'.repeat(8 - account.length)}${account}`, 'hex');
+};
+
+const chunkBuffer = (buffer: Buffer, chunkLength: number): ReadonlyArray<Buffer> => {
+  if (buffer.length > chunkLength) {
+    return [buffer.slice(0, chunkLength)].concat(chunkBuffer(buffer.slice(chunkLength), chunkLength));
+  }
+
+  return [buffer];
+};
+
+export class LedgerHandler {
+  public static readonly byteLimit = LedgerTransport.byteLimit;
+  public static readonly type = LedgerTransport.type;
+
+  public static readonly init = async () => {
+    const isSupported = await LedgerTransport.isSupported();
+    if (!isSupported) {
+      throw new LedgerNotSupportedError();
+    }
+    const paths = await LedgerTransport.list();
+    if (paths.length === 0) {
+      throw new LedgerNotDetectedError();
+    }
+    const ledger = await LedgerTransport.open(paths[0]);
+    ledger.setScrambleKey('NEO');
+
+    return new LedgerHandler(ledger);
+  };
+
+  private readonly ledger: HWLedger;
+  private readonly mutableQueue: LedgerQueueRequest[];
+  private mutableRunning: boolean;
+
+  private constructor(ledger: HWLedger) {
+    this.ledger = ledger;
+    this.mutableRunning = false;
+    this.mutableQueue = [];
+  }
+
+  public async getPublicKey(account = 0): Promise<PublicKeyString> {
+    const response = await this.send(Buffer.concat([keyRequestParams, Buffer.from('14', 'hex'), BIP44(account)]));
+
+    return crypto.toECPoint(response).toString('hex');
+  }
+
+  public async sign({ message, account }: { readonly message: Buffer; readonly account: number }): Promise<Buffer> {
+    const data = Buffer.concat([message, BIP44(account)]);
+    const chunks = chunkBuffer(data, 250);
+    const initChunks = chunks.slice(0, chunks.length - 1);
+    const finalChunk = chunks[chunks.length - 1];
+
+    await Promise.all(
+      initChunks.map(async (chunk) => {
+        await this.send(Buffer.concat([initMessageParams, Buffer.from([chunk.length]), chunk]));
+      }),
+    );
+
+    return this.send(
+      Buffer.concat([finalMessageParams, Buffer.from(finalChunk.length.toString(16), 'hex'), finalChunk]),
+    );
+  }
+
+  public async close(): Promise<void> {
+    await this.ledger.close();
+  }
+
+  private async send(msg: Buffer): Promise<Buffer> {
+    // tslint:disable-next-line:promise-must-complete
+    const promise = new Promise<Buffer>((resolve, reject) => {
+      this.mutableQueue.push({
+        msg,
+        resolve,
+        reject,
+      });
+    });
+
+    this.startSendLoop().catch(() => {
+      // do nothing
+    });
+
+    return promise;
+  }
+
+  private async startSendLoop(): Promise<void> {
+    if (this.mutableRunning) {
+      return;
+    }
+
+    this.mutableRunning = true;
+
+    let entry = this.mutableQueue.shift();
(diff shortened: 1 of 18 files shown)


packages/neo-one-client-core/src/__ledger_tests__/LedgerKeyStore.test.ts
```diff
@@ -0,0 +1,184 @@
...
+
+  let keystore: LedgerKeyStore;
+
+  beforeEach(() => {
+    getAccount.mockReset();
+    getAccount.mockImplementation(() => emptyAccount);
+    keystore = new LedgerKeyStore(ledgerProvider);
+  });
+
+  afterEach(() => {
```
dicarlo2: `afterEach(async () => {`

packages/neo-one-client-core/src/__ledger_tests__/LedgerKeyStore.test.ts
```diff
@@ -0,0 +1,184 @@
...
+    keystore.close().catch(() => {
```
dicarlo2: `await keystore.close();`

packages/neo-one-client-core/src/__ledger_tests__/LedgerKeyStore.test.ts
```diff
@@ -0,0 +1,184 @@
+import { utils } from '@neo-one/utils';
```
dicarlo2: make `LedgerKeyStore` take a `LedgerHandler` to use (with a default set to the global `LedgerHandler`) and then pass in a mock for these tests. Move these tests to the regular unit test folder.

I.e. test `LedgerKeyStore` with mocks. If you're also testing `LedgerHandler` functionality here, then it should be moved to the `LedgerHandler` test.

packages/neo-one-client-core/src/user/keystore/LedgerHandler.ts
```diff
@@ -0,0 +1,146 @@
...
+
+    this.mutableRunning = false;
+  }
+
+  private async doSend(msg: Buffer): Promise<LedgerResponse> {
+    const response = await this.ledger.exchange(msg);
+
+    return {
+      message: response.slice(0, response.length - 2),
+      status: response.slice(response.length - 2),
```
dicarlo2: Should throw an error if the status is not successful and instead of returning an object with `message` and `status`, just return the `message`.

packages/neo-one-client-core/src/user/keystore/LedgerHandler.ts
```diff
@@ -0,0 +1,146 @@
...
+    this.ledger = ledger;
+    this.mutableRunning = false;
+    this.mutableQueue = [];
+  }
+
+  public async getPublicKey(account = 0): Promise<PublicKeyString> {
+    const response = await this.send(Buffer.concat([keyRequestParams, Buffer.from('14', 'hex'), BIP44(account)]));
+
+    return crypto.toECPoint(response.message).toString('hex');
+  }
```
dicarlo2: nit: newline

packages/neo-one-client-core/src/user/keystore/LedgerHandler.ts
```diff
@@ -0,0 +1,146 @@
...
+  private mutableRunning: boolean;
+
+  private constructor(ledger: HWLedger) {
+    this.ledger = ledger;
+    this.mutableRunning = false;
+    this.mutableQueue = [];
+  }
+
+  public async getPublicKey(account = 0): Promise<PublicKeyString> {
+    const response = await this.send(Buffer.concat([keyRequestParams, Buffer.from('14', 'hex'), BIP44(account)]));
```
dicarlo2: This should also sign the public key and then verify that the message (i.e. the publickey) is correctly signed. This is what the neon folks are doing since there seems to be a small chance that the ledger returns the incorrect public key.

packages/neo-one-client-core/src/user/keystore/LocalKeyStore.ts
```diff
@@ -387,8 +387,7 @@ export class LocalKeyStore {
 
   private getPrivateKey(id: UserAccountID): BufferString {
     const wallet = this.getWallet({
-      network: id.network,
-      address: id.address,
+      ...id,
```
dicarlo2: revert

packages/neo-one-types/@ledgerhq/hw-transport-u2f/index.d.ts
```diff
@@ -0,0 +1,11 @@
+declare module '@ledgerhq/hw-transport-u2f' {
```
dicarlo2: Are these in definitelytyped?

.vscode/settings.json
```diff
@@ -1,5 +1,5 @@
 {
-  "typescript.tsdk": "node_modules/typescript/lib",
+"typescript.tsdk": "node_modules/typescript/lib",
```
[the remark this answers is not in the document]
danwbyrne: I don't know why this keeps happening 😠

packages/neo-one-client-core/src/__ledger_tests__/LedgerKeyStore.test.ts
danwbyrne: https://jestjs.io/docs/en/api#aftereachfn-timeout

you don't need to async an `afterEach` if I'm understanding this right. Jest awaits any Promises in an afterEach on its own.

packages/neo-one-client-core/src/user/keystore/LedgerHandler.ts
danwbyrne: Could you elaborate? Their [getPublicKey](https://github.com/CityOfZion/neon-wallet/blob/78a43dc05182d0504c0c16107ae92bde2287e21e/app/ledger/neonLedger.js#L100) code seems to be the same as ours.

dicarlo2: oh, never mind, guess they aren't doing that anymore.

danwbyrne:
WOAH IT PASSED!

packages/neo-one-client-core/src/user/keystore/LedgerHandler.ts
```diff
@@ -0,0 +1,134 @@
...
+
+    this.mutableRunning = false;
+  }
+
+  private async doSend(msg: Buffer): Promise<Buffer> {
+    const response = await this.ledger.exchange(msg);
+
+    const status = response.slice(response.length - 2);
+    if (status.toString('hex') !== '9000') {
+      throw new LedgerStatusCodeError(status.toString('hex'));
```
dicarlo2: Let's throw a more specific error for the codes that we know about. For a code that we don't know, we can throw this general error.

dicarlo2: Since this is the only issue, I'm going to go ahead and merge, but please make a followup PR to improve the errors here.

danwbyrne: #524

Issue #540, opened 2018-10-15
neo-one courses package manager


afragapane:
Create package manager for the neo-one courses
-PackageManager class that takes a FileSystem interface and an observable of a package.json content and just manages the dependencies as needed
Can live in neo-one-editor/src/packageManager

https://github.com/CompuIves/codesandbox-client/blob/master/packages/app/src/sandbox/eval/manager.js#L562

PR #555, opened 2018-10-19
Add package manager for courses/browser editor

afragapane:
### Description of the Change
Add package manager to the neo-one courses/browser editor.

### Test Plan
Added unit tests for the PackageManager and Resolver.  Also manually went through cases.

### Applicable Issues
https://github.com/neo-one-suite/neo-one/issues/540


packages/neo-one-editor/src/manager/PackageManager.ts
```diff
@@ -0,0 +1,146 @@
...
+    readonly fs: FileSystem;
+  }) {
+    this.fs = fs;
+    this.dependencies$ = packageJSON$.pipe(
+      distinctUntilChanged(),
+      switchMap(async (packageJSON) =>
+        this.resolveDependencies(packageJSON.dependencies === undefined ? {} : packageJSON.dependencies),
+      ),
+    );
+    this.packages$ = this.dependencies$.pipe(switchMap(async (deps) => this.fetchPackages(deps, 'node_modules')));
```
dicarlo2: Make `pkgPath` be `/node_modules` so it doesn't resolve in the tests to a local directory (and instead, resolves to `/node_modules/....`)

packages/neo-one-editor/src/manager/PackageManager.ts
```diff
@@ -0,0 +1,146 @@
...
+      Object.values(deps).map(async (depInfo) => this.fetchPackage({ depInfo, pkgPath })),
+    );
+
+    return packages.reduce<{ readonly [name: string]: Module }>((acc, pkg) => ({ ...acc, [pkg.name]: pkg }), {});
+  }
+
+  private async fetchPackage({ depInfo, pkgPath }: NMPackageInfo): Promise<Module> {
+    let file;
+    if (depInfo.main !== undefined && depInfo.main.includes('.')) {
+      file = depInfo.main;
```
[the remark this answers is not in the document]
afragapane: getNPMUrl returns either `https://cdn.jsdelivr.net/npm/${name}@${version}` if `file` is undefined or `https://cdn.jsdelivr.net/npm/${name}@${version}/${file}` if `file` is defined.  If no file is included on jsdeliver url, it defaults to giving you the default entry file for that version.  So, this is a check to make sure we still get a file even if the `main` field is not well formed for our purpose.  The reason for this check was bignumber.js.  You can see it's package.json `main` field is just `bignumber`, even though the entry point is actually `bignumber.js`.

packages/neo-one-editor/src/manager/PackageManager.ts
```diff
@@ -0,0 +1,149 @@
...
+    };
+  }
+
+  private async downloadDependency({ name, version, file }: FetchPackageInfo) {
+    try {
+      const pkg = await fetch(this.getNPMUrl({ name, version, file }));
+
+      return pkg.text();
+      // tslint:disable-next-line no-unused
+    } catch (error) {
```
dicarlo2: Remove the `(error)`

packages/neo-one-editor/src/manager/PackageManager.ts
```diff
@@ -0,0 +1,149 @@
...
+  private async writeModules({
+    packages,
+    pkgPath,
+  }: {
+    readonly packages: { readonly [name: string]: Module };
+    readonly pkgPath: string;
+  }) {
+    await Promise.all(
+      Object.entries(packages).map(async ([name, pkg]) => {
+        await this.writeModules({
```
dicarlo2: `Promise.all()` with `writeModule` below

PR #582, opened 2018-10-25
move typescript to worker only

afragapane:
### Description of the Change
Moved `convertFormattingOptions` method from utils to the ts worker.  Updated 2 locations where the method was used.


packages/neo-one-editor/src/monaco/AsyncLanguageService.ts
```diff
@@ -240,6 +240,24 @@ export class AsyncLanguageService {
       ),
     );
 
+  public readonly convertFormattingOptions = (options: monaco.languages.FormattingOptions): ts.FormatCodeOptions => ({
```
dicarlo2: nit: this doesn't need to be a method on the class, it doesn't use any of the class properties. Make it a function.

PR #584, opened 2018-10-25
offload ts calls to worker in wireDiagnostics

afragapane:
### Description of the Change
Title.


packages/neo-one-editor/src/monaco/AsyncLanguageService.ts
```diff
@@ -327,6 +327,11 @@ export class AsyncLanguageService {
   public readonly getEmitOutput = (fileName: string): Promise<ts.EmitOutput> =>
     this.languageService.then((languageService) => languageService.getEmitOutput(fileName));
 
+  public readonly flattenDiagnosticMessageText = (
+    messageText: string | ts.DiagnosticMessageChain,
+    newLine: string,
+  ): string => ts.flattenDiagnosticMessageText(messageText, newLine);
```
dicarlo2: This needs to return a `Promise<string>` - all methods called through a worker are async. It'd be best if the diagnostics are just pre-flattened so we're not calling an async worker method to flatten every diagnostic - i.e. change the calls to `getSyntacticDiagnostics` and `getSemanticDiagnostics` to return the flattened messages directly.

PR #610, opened 2018-10-30
Align text in console

afragapane:
### Description of the Change
Aligned text in TestDetail side of the console.  Also aligned the headers of the two sides of the console for tests.  

image

image


packages/neo-one-editor/src/editor/toolbar/FileText.tsx
```diff
@@ -15,12 +15,14 @@ const LightText = styled(Box)`
   ${prop('theme.fonts.axiformaRegular')};
   ${prop('theme.fontStyles.body1')};
   color: ${prop('theme.gray0')};
+  transform: translate(0px, -2px);
 `;
 
 const DarkText = styled(Box)`
   ${prop('theme.fonts.axiformaRegular')};
-  ${prop('theme.fontStyles.caption')};
+  ${prop('theme.fontStyles.body1')};
```
dicarlo2: This should still be caption (i.e. smaller than the rest of the text)

Issue #570, opened 2018-10-22
Code Coverage (neo-one-node-vm)


danwbyrne:
maybe 2 days, probably 1 but I haven't looked at it yet.

PR #627, opened 2018-11-01
neo-one-node-vm coverage

danwbyrne:
### Description of the Change

Adds coverage for `neo-one-node-vm`, at about 90% coverage. Only things I had issue testing was `opcodes` and `execute`. `Syscalls` and `StackItems` should be at 100% coverage besides a few helper functions in SysCalls and a handful of errors.

### Benefits

Coverageeeee

### Applicable Issues

#570


packages/neo-one-node-vm/src/__tests__/syscalls.test.ts
```diff
@@ -137,6 +173,8 @@ interface TestCase {
   readonly options?: Options;
   // tslint:disable-next-line no-any
   readonly mock?: ((options: { readonly blockchain: any }) => void);
+  readonly error?: string;
+  readonly flags?: Set<string>;
```
dicarlo2: Make the flags a specific type, e.g. `type Flag = 'noPersistingBlock' | 'someOtherFlag'` and here `Set<Flag>`

packages/neo-one-node-vm/src/__tests__/stackItem/StackItem.test.ts
```diff
@@ -273,9 +307,557 @@ describe('StackItem', () => {
...
 
-    // [arr] NIP
+    // [arr] NIPF
     expect(stackItem.decrement()).toEqual(-1);
     // []
     expect(arr.decrement()).toEqual(-4);
   });
 });
+
+describe('Stack Item Tests', () => {
```
dicarlo2: Can you move these into separate files so that they can run in parallel? (I.e. 1 file per stackitem class type)

Issue #703, opened 2018-11-12
Rate limit node requests


dicarlo2:
Use https://www.npmjs.com/package/koa-ratelimit-lru to rate limit requests to the node.

Add to https://github.com/neo-one-suite/neo-one/blob/master/packages/neo-one-node-http-rpc/src/rpcServer%24.ts#L61 (with similar configuration handling with passed in options as the health check options)

PR #735, opened 2018-11-13
feat(node): `tooBusy` and `rateLimiter` middleware additions

danwbyrne:
### Description of the Change

Adds a `tooBusy` middleware to the app router as well as a `rateLimit` middleware. PR'ing so that you can get eyes on what I've done.

Additionally I submitted a PR to DefinitelyTyped for `rateLimit` so that we can keep `@neo-one/types` clear. Also instead of using [`toobusy`](https://github.com/lloyd/node-toobusy) I'm using [`toobusy-js`](https://github.com/strml/node-toobusy) because it already has types, is this an issue?
### Test Plan

???

### Alternate Designs

:thinking:

### Benefits

ratelimiting and denying requests when toobusy are nice for the node.

### Applicable Issues

#703


packages/neo-one-node-http-rpc/src/middleware/tooBusyCheck.ts
```diff
@@ -0,0 +1,17 @@
+import { Context } from 'koa';
+import toobusy_js from 'toobusy-js';
+
+export interface Options {
+  readonly maxLag?: number;
+}
+
+export const tooBusyCheck = (options: Options) => async (ctx: Context, next: () => Promise<void>) => {
+  if (options !== undefined && options.maxLag !== undefined) {
+    toobusy_js.maxLag(options.maxLag);
```
dicarlo2: Set this outside of the request handler (i.e. one lambda up). No reason to be setting this on every request.

Is this the only option available?

packages/neo-one-node-http-rpc/src/rpcServer$.ts
```diff
@@ -70,6 +87,18 @@ export const rpcServer$ = ({
       const rpcMiddleware = rpc({ blockchain, node });
       router.use(context({ monitor }));
 
+      if (rateLimitOptions !== undefined) {
+        const rateLimitMiddleware = rateLimit(rateLimitOptions);
+
+        router.use(rateLimitMiddleware);
```
dicarlo2: Just `router.use(rateLimit(rateLimitOptions))`

packages/neo-one-node-http-rpc/src/rpcServer$.ts
```diff
@@ -70,6 +87,18 @@ export const rpcServer$ = ({
...
+      }
+
+      if (tooBusyCheckOptions !== undefined) {
+        const tooBusyMiddleware = tooBusyCheck(tooBusyCheckOptions);
+
+        router.use(tooBusyMiddleware);
```
dicarlo2: Just `router.use(tooBusyCheck(tooBusyCheckOptions))`

packages/neo-one-node-http-rpc/src/rpcServer$.ts
danwbyrne: cool, I was thinking similar but wasn't sure if we wanted to always be using it.

packages/neo-one-node-http-rpc/src/middleware/tooBusyCheck.ts
```diff
@@ -0,0 +1,38 @@
+import { Context } from 'koa';
+import * as tooBusy from 'toobusy-js';
+
+export interface Options {
+  readonly interval?: number;
+  readonly maxLag?: number;
+  readonly smoothingFactor?: number;
+  readonly onLag?: {
+    readonly fn: (lag: number) => void;
```
dicarlo2: Options can only be JSON, i.e. they can't have functions - remove `onLag`.

packages/neo-one-node-http-rpc/src/rpcServer$.ts
```diff
@@ -70,6 +87,9 @@ export const rpcServer$ = ({
       const rpcMiddleware = rpc({ blockchain, node });
       router.use(context({ monitor }));
 
+      router.use(rateLimit(rateLimitOptions));
```
dicarlo2: There should be a way to disable both of these. Add an optional `disabled` to each of the options and check if it's present.

packages/neo-one-node-http-rpc/src/rpcServer$.ts
```diff
@@ -70,6 +87,14 @@ export const rpcServer$ = ({
       const rpcMiddleware = rpc({ blockchain, node });
       router.use(context({ monitor }));
 
+      if (!(rateLimitOptions !== undefined && rateLimitOptions.disabled)) {
```
dicarlo2: This is really hard to read, generally you want to avoid negating a compound boolean expression. Instead, make this `rateLimitOptions === undefined || !rateLimitOptions.disabled`

packages/neo-one-node-http-rpc/src/rpcServer$.ts
```diff
@@ -70,6 +87,14 @@ export const rpcServer$ = ({
...
+        router.use(rateLimit(rateLimitOptions));
+      }
+
+      if (!(tooBusyCheckOptions !== undefined && tooBusyCheckOptions.disabled)) {
```
dicarlo2: Same here

packages/neo-one-node-http-rpc/src/rpcServer$.ts
danwbyrne: I'll fix them, but my buddy Augustus De Morgan would like I word.

PR #747, opened 2018-11-20
Adding React/None & TypeScript/JavaScript switches to codegen

### LINKED ISSUE #720: Add React codegen configuration switch
Add here: https://github.com/neo-one-suite/neo-one/blob/master/packages/neo-one-server-plugin-project/src/types.ts#L33. Disable by default. Use the flag here: https://github.com/neo-one-suite/neo-one/blob/master/packages/neo-one-server-plugin-project/src/build/generateCommonCode.ts

Find all the places where we expect react code to be generated, e.g. the courses, and make sure we set that flag to true.

davemneo:
### Requirements

Create mechanism for toggling between language (TypeScript/JavaScript) and framework (React / none )

### Description of the Change

Updating schema and making types available.

### Applicable Issues

Resolves #720


davemneo:
As there were no changes, ' yarn cz ' complained that there were no staged files, so I: 
reset my checkout of the branch 
` git reset --soft HEAD^ ` 
` yarn cz ` was then successful

Issue #716, opened 2018-11-12
Graceful transition to new cli/server version


dicarlo2:
When the user runs a cli version that differs from the server:
 - Print a warning if a newer remote version is available
 - Shut down existing server and start new one if the cli version > currently running server version
 - Print an error and exit if the cli version < currently running server version

We can probably relax the last requirement or add a flag like `--force` to allow the user to force a downgrade. We'll also need to store the last running version in case the server is not currently running.

Pointers:
 
 - https://github.com/neo-one-suite/neo-one/blob/master/packages/neo-one-server-client/src/ServerManager.ts - manages the server from a client implementation (e.g. the cli). This should probably throw a generic error that is converted into a message and printed by the cli for the error/warning cases above.

PR #766, opened 2018-11-26
feat(toolchain): show error on mismatched server version

davemneo:
### Requirements

Make client requests aware of server version, when there's a conflict, take appropriate action, see below.

### Description of the Change

 These outcomes are expected both when the server is running or not running:
 * client/server have same version, newer version online
 * client is upgraded, server is not - restart server
 * server is upgraded, client is not - throw version conflict error

### Test Plan
manual testing of the above listed scenarios.

### Benefits
Server will restart with correct version, otherwise user will be made aware of version conflicts. 

### Applicable Issues
RE: #716
Depends on #822


dicarlo2:
Otherwise, lgtm

packages/neo-one-server-client/src/npmCheck.ts
```diff
@@ -0,0 +1,74 @@
...
+}
+const emptyNPMResult: NPMResult = {
+  success: false,
+  version: '',
+  date: 0,
+};
+
+const askNPM = async (pkg: string): Promise<NPMResult> =>
+  new Promise<NPMResult>(async (resolve) => {
+    const infoP = await execa('yarn', ['info', pkg]);
```
dicarlo2: We shouldn't assume that `yarn` or `npm` in particular are installed. Better to just `fetch` the versions from https://yarnpkg.com/ or https://www.npmjs.com/.

packages/neo-one-server-client/src/ServerManager.ts
```diff
@@ -91,6 +74,53 @@ export class ServerManager {
     }
   }
 
+  public async getServerVersion(): Promise<string | undefined> {
+    let versionContents = '';
```
dicarlo2: nit: `let versionContents: string | undefined;`

packages/neo-one-server-client/src/npmCheck.ts
dicarlo2: https://github.com/neo-one-suite/neo-one/blob/master/packages/neo-one-editor-server/src/resolveDependencies.ts#L85 has an example of fetching the package info and resolving the versions.

packages/neo-one-server-client/src/ServerManager.ts
```diff
@@ -91,6 +74,53 @@ export class ServerManager {
...
+      versionContents = await fs.readFile(this.versionPath, 'utf8');
+    } catch (error) {
+      if (error.code === 'ENOENT') {
+        return undefined;
+      }
+      throw error;
+    }
+
+    return versionContents;
+  }
```
dicarlo2: nit: add newline

packages/neo-one-server-client/src/ServerManager.ts
```diff
@@ -166,11 +210,25 @@ export class ServerManager {
     return { pid: child.pid, started: true };
   }
 
-  public async writePID(pid: number): Promise<void> {
+  public async writeVersionPid(pid: number, version: string): Promise<void> {
+    await fs.ensureDir(path.dirname(this.pidPath));
```
dicarlo2: Remove

dicarlo2: Actually, you should be doing `ensureDir` on `path.dirname(this.versionPath)`

dicarlo2: I would just change this to two methods, `writePID` and `writeVersion` and then call both from this method `wirteVersionPID`.

packages/neo-one-server-client/src/ServerManager.ts
```diff
@@ -146,6 +183,13 @@ export class ServerManager {
       return { pid, started: false };
     }
 
+    const npmResult = await npmCheck('@neo-one/server', this.checkPath, CHECK_TIMEOUT_MS, CHECK_DELAY_SEC);
+
+    if (npmResult.success && this.compareVersions(this.serverVersion, npmResult.version)) {
+      /* tslint:disable-next-line: no-console */
+      console.log(` A newer version (${npmResult.version}) of @neo-one/server is available for download `);
```
dicarlo2: This shouldn't assume that logging to console will actually appear to the user. E.g. consider if we build an editor client which makes use of `ServerManager` - this console.log will never appear to the user. Instead, you should signal the fact that there's a newer version via the return value and let the client implementation decide how best to display that to the user.

packages/neo-one-server-client/src/ServerManager.ts
```diff
@@ -146,6 +183,13 @@ export class ServerManager {
       return { pid, started: false };
     }
 
+    const npmResult = await npmCheck('@neo-one/server', this.checkPath, CHECK_TIMEOUT_MS, CHECK_DELAY_SEC);
```
dicarlo2: Should still check even if the server is already alive.

packages/neo-one-server-client/src/npmCheck.ts
davemneo: 👍

packages/neo-one-server-client/src/ServerManager.ts
davemneo: 👍

packages/neo-one-server-client/src/ServerManager.ts
davemneo: 👍

packages/neo-one-server-client/src/ServerManager.ts
davemneo: 👍

packages/neo-one-server-client/src/ServerManager.ts
davemneo: 👍

packages/neo-one-server-client/src/ServerManager.ts
davemneo: 👍

packages/neo-one-cli/src/InteractiveCLI.ts
```diff
@@ -263,8 +263,12 @@ export class InteractiveCLI {
 
             // tslint:disable-next-line no-any
             let spinner: any | undefined;
+            // tslint:disable-next-line no-any
+            let verSpinner: any | undefined;
+
             try {
-              const { pid } = await manager.start({
+              verSpinner = ora(`Checking for @neo-one/server updates...`).start();
```
dicarlo2: This should only show on the `first` iteration, same as the other spinner.

packages/neo-one-cli/src/InteractiveCLI.ts
```diff
@@ -275,6 +279,14 @@ export class InteractiveCLI {
                 },
               });
 
+              if (verSpinner !== undefined) {
+                if (newerVersion === undefined) {
+                  verSpinner.succeed(`@neo-one/server is current`);
+                } else {
+                  verSpinner.fail(`A newer version (${newerVersion}) of @neo-one/server is available for download`);
```
dicarlo2: Is there a warning instead? It's not really a failure for it to be out of date. Also would be good to link to documentation that explains how to update.

packages/neo-one-server-client/src/ServerManager.ts
```diff
@@ -91,6 +74,54 @@ export class ServerManager {
...
+    return versionContents;
+  }
+
+  public async getCurrentVersion({ port }: { readonly port: number }): Promise<string | undefined> {
+    const client = new Client({ port });
+    const startTime = utils.nowSeconds();
+    // tslint:disable-next-line no-loop-statement
+    while (utils.nowSeconds() - startTime <= 5) {
+      try {
+        return client.getVersion();
```
dicarlo2: This needs to actually `await` the returned promise in order to catch errors. I.e.
```
const version = await client.getVersion();

return version;
```

and then ignore the tslint error.

packages/neo-one-server-client/src/ServerManager.ts
```diff
@@ -140,10 +178,15 @@ export class ServerManager {
...
     readonly binary: Binary;
     readonly onStart?: () => void;
-  }): Promise<{ readonly pid: number; readonly started: boolean }> {
+  }): Promise<{ readonly pid: number; readonly started: boolean; readonly newerVersion: string | undefined }> {
+    const npmResult = await npmCheck('@neo-one/server', this.checkPath, CHECK_TIMEOUT_MS, CHECK_DELAY_SEC);
+
+    const newerVersion =
+      npmResult.success && this.compareVersions(this.serverVersion, npmResult.version) ? npmResult.version : undefined;
+
     const pid = await this.checkAlive(port);
```
dicarlo2: this can be done in parallel with `npmCheck`, put it in a `Promise.all`

packages/neo-one-server-client/src/npmCheck.ts
```diff
@@ -0,0 +1,90 @@
...
+  readonly date: number;
+}
+const emptyNPMResult: NPMResult = {
+  success: false,
+  version: '',
+  date: 0,
+};
+
+const askNPM = async (pkg: string): Promise<NPMResult> =>
+  new Promise<NPMResult>(async (resolve) => {
```
dicarlo2: Eliminate this promise

packages/neo-one-server-client/src/npmCheck.ts
```diff
@@ -0,0 +1,90 @@
...
+    const NPM_REGISTRY_URL = 'https://registry.yarnpkg.com';
+    const escapedName = getEscapedNPMName(pkg);
+    const url = `${NPM_REGISTRY_URL}/${escapedName}`;
+
+    await fetch(url).then(async (res) => {
```
dicarlo2: Don't use `.then`

packages/neo-one-server-client/src/npmCheck.ts
```diff
@@ -0,0 +1,90 @@
...
+  new Promise<t>((resolve) => {
+    setTimeout(resolve, delay, {});
+  });
+
+const saveResult = async (file: string, result: NPMResult) => {
+  await fs.writeFile(file, JSON.stringify(result));
+};
+
+const triggerCacheUpdate = async (pkg: string, cachePath: string, timeout: number): Promise<void> => {
+  const result: NPMResult = await Promise.race([askNPM(pkg), promiseTimeout<NPMResult>(timeout)]);
```
dicarlo2: Neat, didn't know `Promise.race` was a standard promise function.

packages/neo-one-server-client/src/npmCheck.ts
```diff
@@ -0,0 +1,90 @@
...
+    version: result.version,
+    date: Date.now(),
+    success: true,
+  });
+};
+
+const readCache = async (file: string) => {
+  try {
+    const contents = await fs.readFile(file, 'utf8');
+    if (contents.length > 0) {
```
dicarlo2: This shouldn't be necessary.

packages/neo-one-server-client/src/npmCheck.ts
```diff
@@ -0,0 +1,90 @@
...
+): Promise<NPMResult> => {
+  const message = await readCache(cachePath);
+
+  if (message === undefined) {
+    return emptyNPMResult;
+  }
+
+  if (Date.now() > message.date + checkDelaySec) {
+    /* tslint:disable-next-line: no-floating-promises */
+    triggerCacheUpdate(pkg, cachePath, timeout);
```
dicarlo2: Add `.catch(() => { // do nothing })` and remove the tslint disable.

packages/neo-one-server-client/src/npmCheck.ts
```diff
@@ -0,0 +1,90 @@
...
+          resolve(npmResult);
+        } catch {
+          // ignore errors...
+        }
+      }
+      resolve(emptyNPMResult);
+    });
+  });
+
+const promiseTimeout = async <t>(delay: number): Promise<t> =>
```
dicarlo2: nit: uppercase `T`

packages/neo-one-server-client/src/npmCheck.ts
```diff
@@ -0,0 +1,90 @@
...
+  new Promise<t>((resolve) => {
+    setTimeout(resolve, delay, {});
```
dicarlo2: nit: we don't usually use this form, rather just use a lambda - `setTimeout(() => resolve({}), delay)`. Also, I don't think you want a generic here, just use `NPMResult`

packages/neo-one-server-client/src/npmCheck.ts
```diff
@@ -0,0 +1,90 @@
...
+
+export const npmCheck = async (
+  pkg: string,
+  cachePath: string,
+  timeout: number,
+  checkDelaySec: number,
+): Promise<NPMResult> => {
+  const message = await readCache(cachePath);
+
+  if (message === undefined) {
```
dicarlo2: Doesn't seem like the cache will ever be filled with this check. I think you can remove this.

packages/neo-one-server-client/src/npmCheck.ts
```diff
@@ -0,0 +1,90 @@
...
+    return emptyNPMResult;
+  }
+
+  if (Date.now() > message.date + checkDelaySec) {
+    /* tslint:disable-next-line: no-floating-promises */
+    triggerCacheUpdate(pkg, cachePath, timeout);
+  }
+
+  return message;
```
dicarlo2: This should probably just return `message.version` with a return type of `Promise<string | undefined>` - returning `undefined` if we don't have a version.

packages/neo-one-server/src/Server.ts
```diff
@@ -145,7 +145,7 @@ export class Server {
               if (pid === undefined) {
                 this.mutableServerDebug.pid = process.pid;
 
-                await manager.writePID(process.pid);
+                await manager.writeVersionPid(process.pid, VERSION);
```
dicarlo2: nit: `writeVersionPID`

packages/neo-one-server-client/src/npmCheck.ts
```diff
@@ -0,0 +1,90 @@
...
+const promiseTimeout = async <t>(delay: number): Promise<t> =>
+  new Promise<t>((resolve) => {
+    setTimeout(resolve, delay, {});
+  });
+
+const saveResult = async (file: string, result: NPMResult) => {
+  await fs.writeFile(file, JSON.stringify(result));
+};
+
+const triggerCacheUpdate = async (pkg: string, cachePath: string, timeout: number): Promise<void> => {
```
dicarlo2: 👍 this is good thinking

packages/neo-one-server-client/src/ServerManager.ts
davemneo: 👍

packages/neo-one-server-client/src/npmCheck.ts
davemneo: After the first run, it should be.  It's never deleted.

packages/neo-one-server-client/src/npmCheck.ts
davemneo: How's this look? https://github.com/neo-one-suite/neo-one/commit/139445b41d98d4625c7dc097100843bc19a6a9f0#diff-42f186175b31114082c92b9cf786a793R17

packages/neo-one-server-client/src/npmCheck.ts
davemneo: 👍

packages/neo-one-cli/src/InteractiveCLI.ts
```diff
@@ -263,18 +263,32 @@ export class InteractiveCLI {
...
+
             try {
-              const { pid } = await manager.start({
+              const { pid, newerVersion } = await manager.start({
                 port,
                 httpPort,
                 binary: createBinary(argv, this.serverConfig),
                 onStart: () => {
                   if (first) {
+                    verSpinner = ora(`Checking for @neo-one/server updates...`).start();
```
dicarlo2: I think this should still be where it was before, you just need to do
```
if (first) {
  verSpinner = ora(`Checking for @neo-one/server updates...`).start();
}
const { pid, newerVersion } = await manager.start({
```

packages/neo-one-server-client/src/npmCheck.ts
dicarlo2: Right, but if there is no file, we return an `emptyNPMResult`, and we never trigger a write to the file.

packages/neo-one-cli/src/InteractiveCLI.ts
```diff
@@ -263,18 +263,32 @@ export class InteractiveCLI {
...
                   }
                 },
               });
 
+              if (verSpinner !== undefined) {
+                if (newerVersion === undefined) {
+                  verSpinner.succeed(`@neo-one/server is current`);
+                } else {
+                  verSpinner.warn(
+                    `A newer version (${newerVersion}) of @neo-one/server is available for download see https://neo-one.io/docs/playground#Updating or we can update the https://github.com/neo-one-suite/neo-one-playground to be self descriptive with installation and update commands as well as a link to the setup docs?`,
```
dicarlo2: This isn't for updating the playground, this is for updating the `neo-one` binary. This should be a section in https://neo-one.io/docs/environment-setup

packages/neo-one-server/src/Server.ts
davemneo: 👍

packages/neo-one-server-client/src/npmCheck.ts
davemneo: Thank you.

packages/neo-one-server-client/src/npmCheck.ts
davemneo: 👍

packages/neo-one-server-client/src/npmCheck.ts
davemneo: 😂

davemneo:
bors try

packages/neo-one-website/docs/0-installation/1-environment-setup.md
```diff
@@ -58,6 +58,14 @@ In addition to the above, if you're developing TypeScript smart contracts using
  - `@neo-one/smart-contract-test` - TypeScript smart contract testing utilitiees.
  - `@neo-one/smart-contract-typescript-plugin` - TypeScript language server plugin for inline compiler diagnostics in your favorite IDE.
 
+### Updating
```
dicarlo2: Looks like some extra changes slipped in.

packages/neo-one-website/docs/0-installation/2-playground.md
```diff
@@ -15,7 +15,8 @@ This guide will walk you through getting started with the [NEO•ONE Playground]
 ## Requirements
 
   - [Node](https://nodejs.org) >= 8.9.0 (We recommend the latest version)
-  - a package manager: [yarn](https://yarnpkg.com/)   **OR**   npm (distributed with Node)
+  - [git](https://git-scm.com/)
```
dicarlo2: Here too.

packages/neo-one-website/docs/0-installation/1-environment-setup.md
```diff
@@ -58,6 +58,13 @@ In addition to the above, if you're developing TypeScript smart contracts using
...
+
+```bash
+#npm update @neo-one/server
+yarn upgrade @neo-one/server
```
dicarlo2: This should be `yarn upgrade @neo-one/cli`

dicarlo2: Now that I think about it - everywhere in this PR where we use `@neo-one/server` should actually be `@neo-one/cli`. Users will likely never install `@neo-one/server` directly, it's a transitive dependency of `@neo-one/cli` which they will install. E.g. the spinner which says that it's checking for `@neo-one/server` updates will just confuse users since they never installed `@neo-one/server` themselves.

davemneo: 👍

davemneo:
bors try

davemneo:
bors try

davemneo:
bors try

packages/neo-one-website/docs/0-installation/1-environment-setup.md
dicarlo2: `@neo-one/cli`

davemneo:
bors try

davemneo:
bors try

davemneo:
bors p=5

PR #785, opened 2018-11-29
feat(toolchain): e2e tests for JavaScript codegen

davemneo:
In testing phase... 
even with a package.json specifying contract and generated paths as well as codegen framework & language, and then doing...

yarn build
yarn neo-one reset --static-neo-one && yarn neo-one start server --static-neo-one --debug
yarn neo-one build

the resulting output still has ES6 style imports
added some "throw new error" that are not showing up in either steam

### Requirements

Support compiling to JavaScript

### Description of the Change

updating webpack configuration to compile to JavaScript

### DIFF
--- a/package.json
+++ b/package.json
@@ -251,7 +251,7 @@
     "webpack-bundle-analyzer": "3.0.3",
     "webpack-node-externals": "1.7.2",
     "webpack-serve": "2.0.3",
-    "webpackbar": "3.1.3",
+    "webpackbar": "3.1.4",
     "workbox-webpack-plugin": "3.6.3",
     "yargs": "12.0.5"
   },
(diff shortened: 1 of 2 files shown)


packages/neo-one-server-plugin-project/src/__e2e__/buildCommandJs.test.ts
```diff
@@ -0,0 +1,459 @@
+import {
```
dicarlo2: Can you extract out the common logic between this file and `buildCommand.test.ts`

davemneo: There's lots of "any" in this solution: https://github.com/neo-one-suite/neo-one/pull/785/files#diff-28076ab6b09c780f81fa81b6b75c4aaaR11

davemneo: 👍

Issue #797, opened 2018-12-03
update resolveDependencies usage of LRUCache


davemneo:
Update shared failing unit test: resolveDependencies.test.ts to address proper usage of LRUCache.
![image.png](https://zube.io/files/neo-one/f659c37f90d26b23bbb496ea9cfee02d-image.png)

PR #798, opened 2018-12-03
fix(resolveDependencies): Update resolveDependencies to use 'new' when calling LRUCache constructor

davemneo:
### Requirements

update resolveDependencies to call LRUCache constructor with "new"

### Test Plan

    yarn jest resolveDependencies

### Applicable Issues

#797


davemneo:
already resolved.

PR #803, opened 2018-12-04
refactor(getEscapedName): to neo-one-utils-node

davemneo:
### Description of the Change

Moving to shared package
 

# SEE https://github.com/neo-one-suite/neo-one/pull/822


davemneo:
SEE #822

Issue #702, opened 2018-11-12
Use dgeni to create route data for reference pages


dicarlo2:
https://github.com/angular/dgeni

Will require some experimentation and custom work since it's designed to output documentation directly, whereas we just want the processed typescript documentation as a json blob to pass as route data. E.g. we might need to output files as a dummy template using dgeni and then read those files in to pass to the route data (i.e. wasted effort writing to file).

PR #849, opened 2018-12-17
docs(Reference): Rearrange sc doc file for better dgeni parsing.

afragapane:
The order in which same-named items are parsed matters for dgeni. This ordering preserves all the
data we need.

#702


packages/neo-one-smart-contract/src/index.d.ts
```diff
@@ -1404,7 +1403,7 @@ export type Hashable = number | string | boolean | Buffer;
 /**
  * Contains various cryptography functions.
  */
-export interface Crypto {
+export interface CryptoConstructor {
```
dicarlo2: Changing this name broke the tests, you're going to have to track down where we're using `'Crypto'` by name within the compiler and change it to `'CryptoConstructor'`

Issue #710, opened 2018-11-12
Automate data backups


dicarlo2:
Setup a public google cloud bucket for storing the backups. Setup a circleci cron job with the node image that downloads the latest backup, synchronizes the blockchain to the current time, re-uploads the backup, cleans up 1 backup (we should keep say 7 days worth) and also points/copies the latest file to a file named with the suffix `latest`. It should also create a chain dump file. Pointers:

 - https://github.com/neo-one-suite/neo-one/tree/master/packages/neo-one-node-data-backup - contains the code for backing up and restoring from google cloud.
 - https://github.com/neo-one-suite/neo-one/blob/master/packages/neo-one-node/src/fullNode%24.ts#L226 example usage in the node for backing up on a schedule.
 - https://github.com/neo-one-suite/neo-one/tree/master/packages/neo-one-node-offline contains the code for dumping a chain file.

We should just go ahead and use the same entry point file (https://zube.io/neo-one/neo-one/c/714) and simply add additional configuration that controls the "mode" we're running in - i.e. the normal continuous mode or the above mode where we shutdown after finishing the work we need to complete.

danwbyrne:
When it is restoring it will just sit there with no output. Would be nice to know it isn't just hanging.

danwbyrne:
breaking: temporarily disabling `@neo-one/node-data-backup/src/provider/extract.ts` in `@neo-one/node-data-backup/src/provider/GCloudProvider.ts` and using `tar.extract(...)` on each downloaded backup instead of using the `fileStream` approach. This means there is no `Throttle` on the bytes written per second by the extraction method, `writeBytesPerSecond` will be unused.

PR #866, opened 2018-12-20
feat(node): Add Dockerfile and node-config for automatic data backups

danwbyrne:
### Description of the Change

Small alterations to `restore` and `sync` processes in the Node to configure for automatic data backups. Adds `scripts/node/backup.config.json` for the process. I have the credentials in the folder *locally* but I `.gitignored` it. 

Because I wasn't sure if we wanted to set up this as another automatic build I didn't want to just base it off the latest image from `quay` so its just an altered `DockerfileNode` which moves over the config and runs it.

breaking: temporarily disabling @neo-one/node-data-backup/src/provider/extract.ts in @neo-one/node-data-backup/src/provider/GCloudProvider.ts and using tar.extract(...) on each downloaded backup instead of using the fileStream approach. This means there is no Throttle on the bytes written per second by the extraction method, writeBytesPerSecond will be unused.

several interfaces (such as GCloudProviderOptions) have writeBytesPerSecond as a required value, so I will continue defining it in configs until we decide what we are doing about it.

### Test Plan

I have been testing it pretty rigorously, but if you'd like to try then do:

```
cd neo-one
docker build -t test/backup -f scripts/DockerfileBackup .
docker run test/backup
```

this should run the whole process (for me it takes ~1 hour in total).

### Alternate Designs

Persisted data volume between runs? Wouldn't have to restore 15gb+ of data every time.

### Benefits

BLOCKCHAIN SYNC

### Possible Drawbacks

see breaking change above.

### Applicable Issues

#710


packages/neo-one-node-data-backup/src/provider/GCloudProvider.ts
```diff
@@ -3,8 +3,9 @@ import { File } from '@google-cloud/storage/build/src/file';
 import { Monitor } from '@neo-one/monitor';
 import * as fs from 'fs-extra';
 import * as path from 'path';
+import * as tar from 'tar';
 import { Environment } from '../types';
-import { extract } from './extract';
+// import { extract } from './extract';
```
dicarlo2: This should still use `extract`, just replace the implementation with just `tar.extract`

packages/neo-one-node-protocol/src/Node.ts
```diff
@@ -717,7 +724,7 @@ export class Node implements INode {
       connected
         .map((peer) => {
           const { address, port } = peer;
-          const host = new Address6(address);
+          const host = Address6.fromAddress4(address);
```
dicarlo2: Fix this in a separate PR

packages/neo-one-node-protocol/src/Node.ts
```diff
@@ -206,6 +210,8 @@ export class Node implements INode {
       }
 
       this.requestBlocks();
+    } else if (peer !== undefined && this.mutableHaltOnSync) {
+      this.halt$.next(true);
```
dicarlo2: Change this to just a callback that you thread through, it'll be a lot more clear than trying to pass it back up through the observable. So just call something here like `this.onSynchronized`

dicarlo2: Also, I'm not sure this is the right place to do this. What happens if `peer` is `undefined` or `block.index > peer.data.startHeight`? (line 188)

dicarlo2: Take a look at how we check if the node is "ready" (i.e. fully synchronized) here: https://github.com/neo-one-suite/neo-one/blob/master/packages/neo-one-node-http-rpc/src/middleware/checkReady.ts. Should probably just do that on an interval until it's true.

scripts/DockerfileBackup
```diff
@@ -0,0 +1,65 @@
+FROM neotracker/neo-one-circleci-node:10.13.0-1 as builder
```
dicarlo2: I don't think we need a separate docker file for this - the credentials will just be provided in the environment that the docker image is running in.

scripts/node/backup.config.json
```diff
@@ -0,0 +1,26 @@
+{
```
dicarlo2: Similarly, the options for where to backup to should be provided as arguments or by the environment that the image is running in (via env variables in kubernetes or the equivalent in docker compose).

packages/neo-one-node-protocol/src/Node.ts
danwbyrne: > Also, I'm not sure this is the right place to do this. What happens if `peer` is `undefined` or `block.index > peer.data.startHeight`? (line 188)

My thought process here was:
- we check in both `if`s that peer !== undefined
- this peer, since its defined would be the *best* since we define it as:
```
const peer = this.mutableBestPeer;
```
- if our best peer is lower index than us, then we must have just lost connection to the `true` best peer we requested from?
- if our best peer is equal to us then we're ready.

danwbyrne: > Take a look at how we check if the node is "ready" (i.e. fully synchronized) here: https://github.com/neo-one-suite/neo-one/blob/master/packages/neo-one-node-http-rpc/src/middleware/checkReady.ts. Should probably just do that on an interval until it's true.

Where would this interval be taking place?

packages/neo-one-node-data-backup/src/provider/GCloudProvider.ts
danwbyrne: Should I just delete all of the stream extract stuff? I was doing this as a temporary disable if we wanted to come back to fixing the `Throttle`. I guess it works tho so maybe we don't even need the throttle now. 🤔

packages/neo-one-node-protocol/src/Node.ts
danwbyrne: > Change this to just a callback that you thread through, it'll be a lot more clear than trying to pass it back up through the observable. So just call something here like `this.onSynchronized`

so instead of passing it up the observable, I would just pass my FullNode *halted* `stop` function: ```
public async stop(halted?: boolean): Promise<void> {
    if (this.mutableSubscription !== undefined) {
      this.mutableSubscription.unsubscribe();
      this.mutableSubscription = undefined;
      await finalize.wait();

      if (halted) {
        const options = await this.options.options$.pipe(take(1)).toPromise();
        const backupOptions = options.backup !== undefined ? options.backup.options : undefined;
        if (backupOptions !== undefined) {
          await this.backup(backupOptions);
        }
      }
    }
  }
```
to the Node itself to call?

File: packages/neo-one-node-data-backup/src/provider/GCloudProvider.ts (hunk 1 above)
Comment (danwbyrne) [reply packages/neo-one-node-data-backup/src/provider/GCloudProvider.ts]: Also where does this leave `writeBytesPerSecond`? referring to my *breaking* comment.

File: scripts/node/backup.config.json (hunk 5 above)
Comment (danwbyrne) [reply scripts/node/backup.config.json]: With that note, should I remove the default CMD in the current `DockerfileNode`? Seems like it would be better that way if these will ultimately just be used in a k8/compose cluster and not as a one off.

File: packages/neo-one-node/src/FullNode.ts [hunk 6]
```diff
@@ -34,7 +38,23 @@ export class FullNode {
             reject(error);
           }
         };
-        this.mutableSubscription = fullNode$(this.options).subscribe({
+        this.mutableSubscription = combineLatest(
```
danwbyrne: I'm thinking `merge` is more applicable here?

packages/neo-one-node-data-backup/src/provider/extract.ts
```diff
@@ -1,50 +1,17 @@
-import * as fs from 'fs';
-import { Throttle } from 'stream-throttle';
 import * as tar from 'tar';
 
 export const extract = async ({
   downloadPath,
   dataPath,
+  // tslint:disable-next-line:no-unused
```
dicarlo2: Remove this and the next line.

packages/neo-one-node/src/FullNode.ts
dicarlo2: I would just subscribe to the `interval` observable separately - `start` should resolve once the node has started, and trying to combine them seems problematic.

packages/neo-one-node/src/FullNode.ts
```diff
@@ -34,7 +38,23 @@ export class FullNode {
...
+          fullNode$(this.options),
+          this.options.environment.haltOnSync
+            ? interval(5000).pipe(
+                map(async () => {
+                  try {
+                    const response = await fetch(`http://localhost:${port}/ready_health_check`);
+                    if (response.status === 200) {
+                      await this.stop(true);
```
dicarlo2: Move this into it's own `switchMap` otherwise there's a race condition. (Try to figure out why)

packages/neo-one-node/src/FullNode.ts
```diff
@@ -34,7 +38,23 @@ export class FullNode {
             reject(error);
           }
         };
-        this.mutableSubscription = fullNode$(this.options).subscribe({
+        this.mutableSubscription = combineLatest(
+          fullNode$(this.options),
+          this.options.environment.haltOnSync
+            ? interval(5000).pipe(
+                map(async () => {
```
dicarlo2: `switchMap`

packages/neo-one-node/src/fullNode$.ts
```diff
@@ -38,12 +38,14 @@ interface TelemetryEnvironment {
 export interface Environment {
   readonly dataPath: string;
   readonly rpc: RPCServerEnvironment;
+  readonly haltOnSync?: boolean;
```
dicarlo2: `haltAndBackup` is probably a better name for this.

scripts/DockerfileNode
```diff
@@ -59,5 +60,6 @@ RUN yarn build:node
 
 FROM node:10-stretch-slim
 COPY --from=builder tmp/neo-one/dist/ /tmp/
+COPY --from=builder tmp/neo-one/scripts/ /tmp/neo-one/
```
dicarlo2: Remove

packages/neo-one-node/src/FullNode.ts
```diff
@@ -52,12 +78,27 @@ export class FullNode {
...
+
+    if (halted) {
+      const options = await this.options.options$.pipe(take(1)).toPromise();
+      const backupOptions = options.backup !== undefined ? options.backup.options : undefined;
+      if (backupOptions !== undefined) {
+        await this.backup(backupOptions);
+      }
+    }
+
+    if (this.mutableWatcher !== undefined) {
```
dicarlo2: Move this above the other subscription unsubscribe.

packages/neo-one-node/src/FullNode.ts
```diff
@@ -52,12 +78,27 @@ export class FullNode {
     return Promise.resolve();
   }
 
-  public async stop(): Promise<void> {
+  public async stop(halted?: boolean): Promise<void> {
+    this.mutableWatchStarted = false;
```
dicarlo2: Remove, I don't think you need this, just unsubscribe.

scripts/node/backup.config.json
```diff
@@ -0,0 +1,39 @@
...
```
dicarlo2: Remove, no?

packages/neo-one-node-bin/src/utils/getConfiguration.ts
```diff
@@ -67,6 +69,8 @@ const getFreshConfig = (): NodeBinEnvironment => {
     settingsEnvironment: config.settings,
     monitorEnvironment: config.monitor,
     options: config.options === undefined ? {} : config.options,
+    dumpChainFile: config.dumpChainFile,
```
dicarlo2: We should be uploading this chain file somewhere.

danwbyrne: I just submitted a change so that the dumpChainFile would stop crashing because of memory overflows. I thought it was an issue of the writestream not draining, but actually it was the 
```
const blocks = await Promise.all(chunk.map(async (index) => blockchain.block.get({ hashOrIndex: index })));
```
line creating too big an array. I reduced this to only poll 1000 blocks at a time instead of 10000 which solved the problem and had no impact on the processing speed.

Should I just configure them to upload alongside the backup?

PR #899, opened 2019-01-03
build(gulpfile): Add custom transforms for rxjs imports

afragapane:
### Description of the Change
Move from gulp babel transforms to tsc custom transforms to transform rxjs imports.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -22,6 +22,227 @@ const rollup = require('rollup');
 const rollupString = require('rollup-plugin-string');
 const rollupTypescript = require('rollup-plugin-typescript2');
 
+const rxjsTypes = new Set(['Observer']);
+
+// Reference: https://github.com/browserify/browserify#compatibility
+const browserifyModules = {
+  assert: {
+    name: 'assert',
+    version: '1.4.1',
+  },
+  buffer: {
+    name: 'buffer',
+    version: '5.2.1',
+  },
+  console: {
+    name: 'console-browserify',
+    version: '1.1.0',
+  },
+  constants: {
+    name: 'constants-browserify',
+    version: '1.0.0',
+  },
+  crypto: {
+    name: 'crypto-browserify',
+    version: '3.12.0',
+  },
+  domain: {
+    name: 'domain-browser',
+    version: '1.2.0',
+  },
+  events: {
+    name: 'events',
+    version: '3.0.0',
+  },
+  os: {
+    name: 'os-browserify',
+    version: '0.3.0',
+  },
+  path: {
+    name: 'path-browserify',
+    version: '1.0.0',
+  },
+  punycode: {
+    name: 'punycode',
+    version: '2.1.1',
+  },
+  querystring: {
+    name: 'querystring-es3',
+    version: '0.2.1',
+  },
+  stream: {
+    name: 'stream-browserify',
+    version: '2.0.2',
+  },
+  string_decoder: {
+    name: 'string_decoder',
+    version: '1.2.0',
+  },
+  timers: {
+    name: 'timers-browserify',
+    version: '2.0.10',
+  },
+  tty: {
+    name: 'tty-browserify',
+    version: '0.0.1',
+  },
+  url: {
+    name: 'url',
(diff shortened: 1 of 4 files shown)


davemneo:
bor r-

gulpfile.js
```diff
@@ -22,6 +23,115 @@ const rollup = require('rollup');
...
+
+  return statements;
+};
+
+const createCustomTransform = (context) => {
+  return (node) => {
+    const visitor = (node) => {
+      if (typescript.isSourceFile(node)) {
+        let statements = node.statements;
+        statements = transformImport(`rxjs`, transformRxjsImportText, statements);
```
dicarlo2: The way you're doing this is `O(n*m)` complexity, where `n` is the number of statements in the file and `m` is the number of imports to map. Please change it to `O(n)`. I'll leave it as an exercise to figure out how :)

Note, I fully realize it's probably not an issue here, but it's good practice to always use the most optimal solution.

gulpfile.js
```diff
@@ -22,6 +23,115 @@ const rollup = require('rollup');
...
+          node.typeReferences,
+          node.hasNoDefaultLib,
+          node.libReferences,
+        );
+      }
+
+      return node;
+    };
+
+    typescript.visitEachChild(node, visitor, context);
```
dicarlo2: I don't think you need this.

gulpfile.js
```diff
@@ -22,6 +23,115 @@ const rollup = require('rollup');
...
+            node.importClause,
+            typescript.createStringLiteral(`${moduleName}-browserify`),
+          );
+        }
+      }
+
+      return node;
+    };
+
+    typescript.visitEachChild(node, visitor, context);
```
dicarlo2: Ditto

gulpfile.js
```diff
@@ -22,6 +23,115 @@ const rollup = require('rollup');
...
+    const visitor = (node) => {
+      if (typescript.isImportDeclaration(node)) {
+        const moduleName = node.moduleSpecifier.text;
+        if (isBuiltinModule(moduleName)) {
+          return typescript.updateImportDeclaration(
+            node,
+            node.decorators,
+            node.modifiers,
+            node.importClause,
+            typescript.createStringLiteral(`${moduleName}-browserify`),
```
dicarlo2: You need to check that all of these modules actually exist. (Hint: they don't, so we should only be transforming the ones that do)

tsconfig/tsconfig.es2017.browserify.cjs.json
```diff
@@ -0,0 +1,14 @@
+{
```
dicarlo2: I don't think you need these browserify configs.

tsconfig.json
```diff
@@ -34,6 +34,7 @@
     "noFallthroughCasesInSwitch": true,
     "forceConsistentCasingInFileNames": true
   },
+  "include": ["**/*"],
```
dicarlo2: Remove, shouldn't be necessary (this is the default)

packages/neo-one-worker/package.json
```diff
@@ -8,5 +8,25 @@
     "rxjs": "^6.3.3",
     "tslib": "^1.9.3"
   },
-  "sideEffects": false
+  "sideEffects": false,
+  "browserify": {
```
dicarlo2: Remove?

gulpfile.js
```diff
@@ -22,6 +22,224 @@ const rollup = require('rollup');
...
+      return node;
+    };
+
+    typescript.visitEachChild(node, visitor, context);
+
+    return typescript.visitNode(node, visitor);
+  };
+};
+
+let packageBrowserifyMap = {};
```
dicarlo2: The gulp steps run in parallel, so you need to make sure the transform package.json always runs after the typescript transforms. Also need to verify that everything runs in the same process. Did you check that the output has the expected dependencies?

dicarlo2: Ah, I see that you did and it does run afterwards now. Cool.

dicarlo2:
Looks like e2e tests are failing though.

PR #902, opened 2019-01-03
fix(node): add a catch for better address resolution

danwbyrne:
### Description of the Change

Sometimes `new Address6(address)` doesn't get an actual address6, so we need to use the static `Address6.fromAddress4(address)` when that happens.

### Test Plan

passing in these RPC URLs ```
[
    "http://seed6.ngd.network:10332",
    "http://seed10.ngd.network:10332",
    "http://node1.sgp1.bridgeprotocol.io:10332",
    "http://seed1.aphelion-neo.com:10332",
    "https://seed1.red4sec.com:10332"
]
```
used to fail, but now they are resolved using the second option.

### Alternate Designs

We could try the static method first? 

### Benefits

RPC URLs resolve properly.


davemneo:
bors try

davemneo:
bors try

PR #941, opened 2019-01-15
fix(build): adjust several packages to allow emitting .d.ts files.

danwbyrne:
### Description of the Change

- temporarily use custom `bn.js` type definition which uses `export default`
- `export` a few local interfaces/functions that need to be exposed for declaration gen.
- `** @internal *` mark `TransactionBase` to hide the mixin function from declaration.

While we use the temporary `bn.js/index.d.ts` I have opened in issue in definitelyTyped:
https://github.com/DefinitelyTyped/DefinitelyTyped/pull/32178

### Test Plan

```
yarn install
yarn build
```
`/dist` should now *only* have `.d.ts` and `.js` files.

### Alternate Designs

I think the only alternative is to handwrite the declaration files.

### Benefits

Should fix compiling a project which uses our packages but utilizes a different `tsconfig`.

### Possible Drawbacks

Need eyes on the `gulpfile.js` changes, not *wholly* confident in what I've done but essentially I just removed copying the typescript files to dist and compiling in place, in favor of compiling from src to dist. Do my changes implicate missing transforms? 🤷‍♂️

### Issues

#951

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -124,22 +124,30 @@ const globs = {
     '!packages/*/src/__e2e__/**/*',
     '!packages/*/src/bin/**/*',
   ].concat(skipGlobs),
-  src: (format) => [
-    `${getDistBase(format)}/packages/*/src/**/*.{ts,tsx}`,
-    `!${getDistBase(format)}/packages/*/src/**/*.test.{ts,tsx}`,
-    `!${getDistBase(format)}/packages/*/src/__data__/**/*`,
-    `!${getDistBase(format)}/packages/*/src/__tests__/**/*`,
-    `!${getDistBase(format)}/packages/*/src/__e2e__/**/*`,
-    `!${getDistBase(format)}/packages/*/src/bin/**/*`,
-    `!${getDistBase(format)}/packages/neo-one-developer-tools/src/*.ts`,
-    `!${getDistBase(format)}/packages/neo-one-developer-tools-frame/src/*.ts`,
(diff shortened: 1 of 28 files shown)


tsconfig.json
```diff
@@ -18,7 +18,8 @@
     },
     "pretty": true,
     "noEmit": true,
-    "declaration": false,
+    "declaration": true,
```
dicarlo2: I think these 2 should just be in the tsconfig/ directory files. The top-level config (this file) is meant just for type checking and the editor to use.

dicarlo2:
Actually looks like the test failures are legit.

danwbyrne:
The resolution of my pull request to DefinitelyTyped was not in our favor so I think we will have to continue using our own definition of `bn.js`. The issue being that it is a breaking change to libraries that don't use `esModuleInterop: true`. They suggested hacky ways to fix the errors but none are real solutions. There may be a cleaner way to 're-export' the declaration from the up-to-date types as default, but I haven't gotten that to work so far.

Let me know where you think we should go from here with this @dicarlo2 , I've gotten the build flattened so there is no `src` in the module directories, I'll be looking at the node for the rest of the day. Also when comparing across builds (this one, and the current one) it looks like I'm compiling more than I should? If you find time I'd like to talk about expected build output with you.

dicarlo2:
LGTM, just need to fix the conflicts.

danwbyrne:
bors p=1

danwbyrne:
bors plz

Issue #751, opened 2018-11-20
Investigate tests causing 1000ms+ lag and triggering the new `tooBusy` middleware


danwbyrne:
as of [neo-one-suite/neo-one#710 (neo-one-suite/neo-one#704)](https://zube.io/neo-one/neo-one/c/710) multiple tests fail the default 70ms tolerance, however a lot of them cause a suspicious amount of lag (over 1000ms). Need to resolve. Temporarily upped test default to 1000ms.

davemneo:
What are the reproduction steps?

dicarlo2:
TODO: Test with prod infra + https://artillery.io/

Issue #750, opened 2018-11-20
Investigate tests hitting the 2000/minute request limit


danwbyrne:
As of [neo-one-suite/neo-one#709 (neo-one-suite/neo-one#703)](https://zube.io/neo-one/neo-one/c/709) the new `rateLimit` middleware e2e tests `bootstrap20Wallets` fails because it hits the rateLimit. Increased it for now until this is resolved.

davemneo:
Is this the failing test? 

![image.png](https://zube.io/files/neo-one/ab14c9f316d2d39d11d2c890002e45e5-image.png)

PR #986, opened 2019-01-25
fix(rpcServer): Update configuration for rateLimit and tooBusy

davemneo:
### Requirements

   Fix rateLimit and tooBusy calls so as to allow for reasonable limits when running unit tests

### Description of the Change
Increased rateLimit
Added a static maxLag for tooBusy

### Test Plan
    yarn jest

### Possible Drawbacks
May introduce a delay in rpcServer calls 

### Applicable Issues
fix: #750, fix: #751

### DIFF
--- a/packages/neo-one-node-http-rpc/src/middleware/tooBusyCheck.ts
+++ b/packages/neo-one-node-http-rpc/src/middleware/tooBusyCheck.ts
@@ -8,15 +8,19 @@ export interface Options {
   readonly smoothingFactor?: number;
 }
 
-export const tooBusyCheck = (options: Options) => {
-  if (options.interval !== undefined) {
-    tooBusy.interval(options.interval);
-  }
-  if (options.maxLag !== undefined) {
-    tooBusy.maxLag(options.maxLag);
-  }
-  if (options.smoothingFactor !== undefined) {
-    tooBusy.smoothingFactor(options.smoothingFactor);
+tooBusy.maxLag(5000);
+
+export const tooBusyCheck = (options?: Options) => {
+  if (options !== undefined) {
+    if (options.interval !== undefined) {
+      tooBusy.interval(options.interval);
+    }
+    if (options.maxLag !== undefined) {
+      tooBusy.maxLag(options.maxLag);
+    }
+    if (options.smoothingFactor !== undefined) {
+      tooBusy.smoothingFactor(options.smoothingFactor);
+    }
   }
 
   return async (ctx: Context, next: () => Promise<void>) => {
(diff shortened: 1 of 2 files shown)


packages/neo-one-node-http-rpc/src/middleware/tooBusyCheck.ts
```diff
@@ -8,15 +8,19 @@ export interface Options {
...
-export const tooBusyCheck = (options: Options) => {
-  if (options.interval !== undefined) {
-    tooBusy.interval(options.interval);
-  }
-  if (options.maxLag !== undefined) {
-    tooBusy.maxLag(options.maxLag);
-  }
-  if (options.smoothingFactor !== undefined) {
-    tooBusy.smoothingFactor(options.smoothingFactor);
+tooBusy.maxLag(5000);
```
davemneo: With much testing, this was found to be the sweet spot for tests to run successfully.

packages/neo-one-node-http-rpc/src/rpcServer$.ts
```diff
@@ -105,12 +105,12 @@ export const rpcServer$ = ({
         router.get(readyMiddleware.name, readyMiddleware.path, readyMiddleware.middleware);
       }
 
-      if (rateLimitOptions !== undefined && rateLimitOptions.enabled) {
-        router.use(rateLimit(rateLimitOptions));
+      if (rateLimitOptions === undefined || rateLimitOptions.enabled) {
+        router.use(rateLimit({ rate: 6000 }));
```
davemneo: With much testing, this was found to be the sweet spot for tests to run successfully

packages/neo-one-node-http-rpc/src/rpcServer$.ts
```diff
@@ -105,12 +105,12 @@ export const rpcServer$ = ({
         router.get(readyMiddleware.name, readyMiddleware.path, readyMiddleware.middleware);
       }
 
-      if (rateLimitOptions !== undefined && rateLimitOptions.enabled) {
```
dicarlo2: The changes to these conditions should be reverted, no?

packages/neo-one-node-http-rpc/src/rpcServer$.ts
dicarlo2: Is this 6000 rps? That seems excessive - which tests are causing this to trip?

packages/neo-one-node-http-rpc/src/middleware/tooBusyCheck.ts
dicarlo2: What's the default here? Which tests are causing this to trip?

dicarlo2:
To be clear, the goal of this is not to just find values that will work - it's to figure out where and why we're tripping reasonable defaults for these values.

danwbyrne:
Lets ice this one and I'll pick it up when I look at the node storage costs? :thinking:

davemneo:
Not finding time to properly prioritize this with current tasks.  Withdrawing this PR.

PR #1004, opened 2019-01-29
fix(website): css-loader minimize option fix

danwbyrne:
### Description of the Change

as of Version 1.0.0, `minimize` is not an option for `css-loader`. https://github.com/webpack-contrib/css-loader/releases/tag/v1.0.0

Based on their recommendations we implement `cssnano`.

### Benefits

`css-loader` error gone from website start.


davemneo:
bors p=4

Issue #1011, opened 2019-01-30
Why are CI e2e tests so bad?


danwbyrne:
CI tests are doo-doo right now, why?

### Steps to Reproduce
examples:

https://circleci.com/gh/neo-one-suite/neo-one/30552
https://circleci.com/gh/neo-one-suite/neo-one/30403
https://circleci.com/gh/neo-one-suite/neo-one/30373?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link ???? (this is from removing appveyor.yml)

**Expected behavior:** e2e tests *run* consistently.

**Actual behavior:** There is like an 8/10 chance one of the above errors happens.

**Reproduces how often:** 50%

### Additional Information

This is just an issue for investigation discussion

I'm curious if it could be related to CI's load during x time of the day.

PR #1022, opened 2019-02-02
test(e2e): Retry starting server with fresh ports.

dicarlo2:
### Requirements

### Description of the Change

It seems that the neo-one server sometimes fails to start because the port is already taken, based on this failure: https://circleci.com/gh/neo-one-suite/neo-one/31184. This change adds retries to starting the server so that a new random port is picked.

### Test Plan

Going to let CI run.

### Alternate Designs

Using a random port for each test is not ideal because of possible collisions, however IIRC I couldn't come up with a scheme for deterministically choosing ports so went ahead with this solution. Adding retries should eliminate the flakiness caused by port collisions.

### Benefits

More consistent e2e tests.

### Possible Drawbacks

Failures to start the server take 1 minute longer. (60 check tries at 1 second each vs 40 check tries at 1 second each times 3 overall tries).

### Applicable Issues

Not sure if one is open.

### DIFF
--- a/scripts/e2e/One.js
+++ b/scripts/e2e/One.js
@@ -21,20 +21,7 @@ class One {
   }
 
   async teardown() {
-    if (this.server === undefined) {
-      return;
-    }
-
-    try {
-      await this._exec('reset --static-neo-one');
-    } catch (error) {
-      this.server.kill('SIGINT');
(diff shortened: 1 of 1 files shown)


danwbyrne:
#1011

davemneo:
bors try

PR #1051, opened 2019-02-07
feat(node): add $PORT as a default for the rpc server

danwbyrne:
### Description of the Change

Lets the node default the PORT environment variable to its `rpc.http.port` value. Necessary for Heroku deploys.

### Benefits

Heroku deploys


danwbyrne:
bors p=1

PR #1054, opened 2019-02-07
fix(token): rename Token to NEP5Token

davemneo:
To allow users to readily identify they are creating a NEP5-Token, the name of the base class will be renamed from Token to NEP5Token.


davemneo:
failing test looks to be related to:
https://github.com/neo-one-suite/neo-one/commit/ecbb23c70534dee29c4c94926f24140cb3590a45

PR #1061, opened 2019-02-08
feat(browser compatibility): Generate a browser-compatible bundle for expo

afragapane:
### Requirements

### Description of the Change

Generate a browser-comaptible bundle for client & client full packages using webpack. All packages
which rely on nodejs builtins are bundled, others are left as external.  Also allow createClient to
be configured with a host so that a private network running on localhost can be accessed from a
phone or emulator. This is all to aid with requested expo compatibility.

### Test Plan
Tested produced bundles in an expo project and observed ability to compile and use neo-one client functionality.

### Benefits
This produces a browser compatible version of the neo-one client for use with tools like expo which do not allow nodejs builtin modules in their bundling process.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -17,94 +17,13 @@ const path = require('path');
 const through2 = require('through2');
 const typescript = require('typescript');
 const pkg = require('./package.json');
+const webpack = require('webpack');
 
 const rollup = require('rollup');
 const rollupString = require('rollup-plugin-string');
 const rollupTypescript = require('rollup-plugin-typescript2');
 
 const rxjsTypes = new Set(['Observer']);
-
-// Reference: https://github.com/browserify/browserify#compatibility
-const browserifyModules = {
-  assert: {
-    name: 'assert',
-    version: '1.4.1',
-  },
-  buffer: {
-    name: 'buffer',
-    version: '5.2.1',
-  },
-  console: {
-    name: 'console-browserify',
-    version: '1.1.0',
-  },
-  constants: {
-    name: 'constants-browserify',
-    version: '1.0.0',
-  },
-  crypto: {
-    name: 'crypto-browserify',
-    version: '3.12.0',
-  },
-  domain: {
-    name: 'domain-browser',
-    version: '1.2.0',
-  },
-  events: {
-    name: 'events',
-    version: '3.0.0',
-  },
-  os: {
-    name: 'os-browserify',
-    version: '0.3.0',
-  },
-  path: {
-    name: 'path-browserify',
-    version: '1.0.0',
-  },
-  punycode: {
-    name: 'punycode',
-    version: '2.1.1',
-  },
-  querystring: {
-    name: 'querystring-es3',
-    version: '0.2.1',
-  },
-  stream: {
-    name: 'stream-browserify',
-    version: '2.0.2',
-  },
-  string_decoder: {
-    name: 'string_decoder',
-    version: '1.2.0',
-  },
-  timers: {
-    name: 'timers-browserify',
-    version: '2.0.10',
-  },
-  tty: {
-    name: 'tty-browserify',
-    version: '0.0.1',
-  },
-  url: {
-    name: 'url',
-    version: '0.11.0',
-  },
-  util: {
-    name: 'util',
-    version: '0.11.1',
-  },
-  vm: {
-    name: 'vm-browserify',
-    version: '1.1.0',
-  },
-  zlib: {
-    name: 'browserify-zlib',
-    version: '0.2.0',
-  },
-};
-
-const modulesToBrowserify = new Set(Object.keys(browserifyModules));
 
 const transformRxjsImportText = (importName) =>
   importName === 'EMPTY'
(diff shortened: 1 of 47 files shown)


packages/neo-one-client-core/src/provider/NEOONEOneDataProvider.ts
```diff
@@ -235,7 +238,7 @@ export class NEOONEOneDataProvider implements DeveloperProvider {
   private async getProvider(): Promise<NEOONEDataProvider> {
     /* istanbul ignore next */
     if (this.mutableProvider === undefined) {
-      const client = new OneClient(this.port);
+      const client = new OneClient(this.port, this.host);
```
dicarlo2: Use `this.host` below to modify the `rpcURL`, something like `result.response.nodes[0].rpcAddress.replace('localhost', this.host)`

packages/neo-one-client-core/src/provider/NEOONEOneDataProvider.ts
```diff
@@ -48,14 +49,16 @@ export class NEOONEOneDataProvider implements DeveloperProvider {
   private mutableProvider: NEOONEDataProvider | undefined;
   private readonly projectID: string;
   private readonly port: number;
+  private readonly host: string | undefined;
   private readonly iterBlocksFetchTimeoutMS: number | undefined;
   private readonly iterBlocksBatchSize: number | undefined;
 
   public constructor(options: NEOONEOneDataProviderOptions) {
-    const { network, projectID, port, iterBlocksFetchTimeoutMS, iterBlocksBatchSize } = options;
+    const { network, projectID, port, host, iterBlocksFetchTimeoutMS, iterBlocksBatchSize } = options;
```
dicarlo2: `host='localhost'`

packages/neo-one-server-plugin-project/src/__data__/ico/one/generated/client.ts
```diff
@@ -118,12 +127,12 @@ export const createClient = <TUserAccountProviders extends UserAccountProviders<
   return new Client(userAccountProviders);
 };
 
-export const createDeveloperClients = (): { [network: string]: DeveloperClient } => ({
-  local: new DeveloperClient(new NEOONEOneDataProvider({ network: 'local', projectID, port: 13319 })),
+export const createDeveloperClients = (host = 'localhost'): { [network: string]: DeveloperClient } => ({
+  local: new DeveloperClient(new NEOONEOneDataProvider({ network: 'local', projectID, host, port: 35179 })),
 });
 
 export const createLocalClients = (): { [network: string]: LocalClient } => {
```
dicarlo2: This needs `host` as well.

gulpfile.js
```diff
@@ -354,12 +176,18 @@ const SKIP_PACKAGES = new Set([
 ]);
 const SKIP_PACKAGES_LIST = [...SKIP_PACKAGES];
 
-const pkgs = fs
-  .readdirSync('packages')
-  .filter((file) => !file.startsWith('.'))
-  .filter((file) => !SKIP_PACKAGES.has(file))
-  .filter((pkg) => fs.pathExistsSync(path.resolve('packages', pkg, 'package.json')));
-const pkgJSONs = pkgs.map((pkg) => [pkg, getPackageJSON(pkg)]);
+const getPackageJSONs = () => {
```
dicarlo2: Why wrap this in a function?

gulpfile.js
```diff
@@ -733,6 +541,110 @@ const copyRootTSConfig = ((cache) =>
...
+
+        webpack(webpackBrowserifyConfig(Object.keys(externals), inputPath, outputPath), (err, stats) => {
+          if (err) {
+            throw err;
+          } else if (stats.hasErrors()) {
+            throw new Error('Webpack bundling failed.');
+          }
+
+          const pkgJSON = fs.readJSONSync(path.resolve(inputPath, 'package.json'));
+          fs.writeJSONSync(
```
dicarlo2: I think you need to make sure all of the other files that are expected in the package are present (i.e. the other ones that are copied over or generated for the other packages)

gulpfile.js
```diff
@@ -733,6 +541,110 @@ const copyRootTSConfig = ((cache) =>
...
+
+const CLIENT_FULL_PACKAGES = CLIENT_PACKAGES.concat(['@neo-one/client-full', '@neo-one/client-full-core']);
+
+const CLIENT_BROWSERIFY = {
+  'neo-one-client': CLIENT_PACKAGES,
+  'neo-one-client-full': CLIENT_FULL_PACKAGES,
+};
+
+const webpackBrowserifyConfig = (externals, inputPath, outputPath) => ({
+  entry: inputPath,
```
dicarlo2: probably want to set `mode: 'production'`

gulpfile.js
```diff
@@ -733,6 +541,110 @@ const copyRootTSConfig = ((cache) =>
...
+    if (format.dist === 'neo-one') {
+      const pkgJSONMap = createPkgJSONMap();
+
+      Object.entries(CLIENT_BROWSERIFY).map(([pkgName, dependencies]) => {
+        const externals = getClientPkgDependencies(pkgJSONMap, dependencies);
+        const inputPath = path.resolve(__dirname, getDest(format), pkgName);
+        const outputPath = path.resolve(__dirname, getDest(format), `${pkgName}-browserify`);
+
+        webpack(webpackBrowserifyConfig(Object.keys(externals), inputPath, outputPath), (err, stats) => {
+          if (err) {
```
dicarlo2: Add
```
      console.log(
        stats.toString({
          performance: false,
          hash: false,
          timings: true,
          entrypoints: false,
          chunkOrigins: false,
          chunkModules: false,
          colors: true,
        }),
      );
```

gulpfile.js
afragapane: So I can call it in my browserify task.  If I don't, then pkgJSONs is empty if I run the task alone.

packages/neo-one-server-plugin-project/src/buildCommand.ts
```diff
@@ -113,15 +114,18 @@ export const buildCommand = ({ cli }: InteractiveCLIArgs) => {
     .option('--no-progress', "Don't output progress. Typically used for CI scenarios")
     .option('--reset', "Reset the local development environment of a NEO•ONE project to it's original state.")
     .option('--watch', 'Watch contracts for changes and rebuild.')
+    .option('--browser', 'Generated code references pre-bundled NEO•ONE client, ready to be used in the browser.')
```
dicarlo2: This needs to be an option in the `ProjectConfig` - https://github.com/neo-one-suite/neo-one/blob/master/packages/neo-one-server-plugin-project/src/types.ts#L36

Otherwise, for example, `neo-one build --reset` doesn't work as expected (see below).

packages/neo-one-server-plugin-project/src/reset/index.ts
```diff
@@ -31,13 +31,15 @@ export const reset = (pluginManager: PluginManager, options: ResetTaskListOption
             const buildOptions: BuildTaskListOptions = {
               command: 'build',
               rootDir: project.rootDir,
+              browser: false,
```
dicarlo2: This is why it needs to be in the project config - we can't assume here that it's always false. E.g. think about what would happen if a user ran `neo-one build --browser` and then later `neo-one build --reset`

gulpfile.js
```diff
@@ -733,6 +540,169 @@ const copyRootTSConfig = ((cache) =>
...
+                chunkModules: false,
+                colors: true,
+              }),
+            );
+            throw new Error('Webpack bundling failed.');
+          }
+
+          neoOneDependencies.forEach((dependency) => {
+            const dirName = getDirName(dependency);
+            copyBrowserifyTypesFiles(path.resolve(basePath, dirName), path.resolve(basePath, `${dirName}-browserify`));
```
dicarlo2: I think you need to copy in the LICENSE and README.md files from the root of the project as well.

dicarlo2: Oh, I see, maybe you already are? Is there a LICENSE and README.md file in the output?

PR #1103, opened 2019-02-17
feat(contract-mixins): Add Pausable

davemneo:
### Requirements

Create a mix-in for Pausable contract.


davemneo:
Currently unable to build contract in test, I've added some console logs to the test.

Possibly related to nested class inheritance, see: https://github.com/neo-one-suite/neo-one/pull/1103/files#diff-db0e6bd7ba5752ace5c640c8ab17b191R6
(or a bad import) 
Files to pay attention to: 
- packages/neo-one-smart-contract-lib/src/__data__/contracts/lifecycle/TestPausable.ts
- packages/neo-one-smart-contract-lib/src/__tests__/lifecycle/TestPausable.test.ts
- packages/neo-one-smart-contract-lib/src/access/index.ts
- packages/neo-one-smart-contract-lib/src/index.ts
- packages/neo-one-smart-contract-lib/src/lifecycle/Pausable.ts

gulpfile.js
```diff
@@ -408,7 +408,7 @@ const copyTypescript = ((cache) =>
       .src(globs.typescript)
       .pipe(
         gulpFilter([
-          'packages/neo-one-smart-contract-lib/src/*.ts',
+          'packages/neo-one-smart-contract-lib/src/**/*.ts',
```
danwbyrne: You should maybe split this change into a separate PR as I think it will affect more than just this one. @davemneo

packages/neo-one-smart-contract-lib/src/__data__/contracts/lifecycle/TestPausable.ts
```diff
@@ -0,0 +1,32 @@
+import { Address, SmartContract, constant } from '@neo-one/smart-contract';
+/* tslint:disable-next-line: no-implicit-dependencies */
```
danwbyrne: Is there a way we can add a *solid* TSLint exclude for these files? Seeing this in every `__data__` file is a bit unintuitive. Not sure if we have that control for single rule ignoring across a glob.

packages/neo-one-smart-contract-lib/src/__tests__/__snapshots__/RedToken.test.ts.snap
```diff
@@ -2,4 +2,4 @@
 
 exports[`RedToken properties + issue + balanceOf + totalSupply + transfer: transfer consume 1`] = `"0"`;
 
-exports[`RedToken properties + issue + balanceOf + totalSupply + transfer: transfer cost 1`] = `"3.389"`;
```
danwbyrne: Curious: does your contract affect the RedToken test? Odd that it makes it cost more gas, no?

tsconfig.json
```diff
@@ -37,16 +37,22 @@
   "exclude": [
     "dist",
     "packages/neo-one-server-plugin-wallet/src/contracts/*.ts",
+    "packages/neo-one-smart-contract-lib/src/access/roles/*.ts",
```
danwbyrne: I think you should just use
```
"packages/neo-one-smart-contract-lib/src/access/**/*.ts,
```
instead of
```
"packages/neo-one-smart-contract-lib/src/access/*.ts",
"packages/neo-one-smart-contract-lib/src/access/roles/*.ts"
```

could probably do the same for lifecycles.

packages/neo-one-smart-contract-lib/src/__tests__/__snapshots__/RedToken.test.ts.snap
davemneo: Shouldn't be.  I'll look into this.  Did any Opcode changes go in?

gulpfile.js
davemneo: I applied it to affected PR's so I don't have to wait for the PR and CI process.

packages/neo-one-smart-contract-lib/src/__data__/contracts/lifecycle/TestPausable.ts
davemneo: @danwbyrne totally agree, but it looks like they don't want to include a Conditional configuration [https://github.com/palantir/tslint/issues/3447]

tsconfig.json
davemneo: done

packages/neo-one-smart-contract-lib/src/__tests__/__snapshots__/RedToken.test.ts.snap
davemneo: This happens because the compiler is grabbing everything that's being exported by `@neo-one/smart-contracts-lib` -- Alex indicated there's a plan to clean this up.

davemneo:
New structure, new PRs to inherit from.

Issue #666, opened 2018-11-07
Use 'ora' instead of 'elegant-spinner'


danwbyrne:
https://github.com/sindresorhus/ora

same author, he links to this repo now from elegant-spinner and says its prefered. Might need to type it, will look into later.

dicarlo2:
https://github.com/neo-one-suite/neo-one/blob/master/packages/neo-one-server-plugin/src/handleCLITaskList.ts#L4 is the only spot where it's used

dicarlo2:
@spencercorwin @danwbyrne is this done?

spencercorwin:
To my knowledge, yes

dicarlo2:
great, thanks!

PR #1127, opened 2019-02-23
Ora spinner

spencercorwin:
### Description of the Change

This removes the elegant-spinner dependency and replaces it with the ora package to implement the spinner used in the CLI, as referenced here: #666. I actually don't think this change is at all necessary. From what I can tell the existing elegant-spinner works well with the CLI handlers. The way ora works is more complicated and could only be properly added to the current CLI handlers by calling `.frame()` on the ora instance so that it returned a string instead of an automatically spinning object. That being said, this PR implements ora perfectly fine and could be extended later. Or it could be ignored.

### Test Plan

I tested this manually by running `yarn neo-one build` in a local Neo-One project and watching/screen recording the CLI with the existing implementation and new implementation. Here are some before and after screenshots to show the spinner working perfectly:

Before:
screen shot 2019-02-22 at 9 43 17 pm 2

screen shot 2019-02-22 at 9 43 36 pm 2

After:
screen shot 2019-02-22 at 9 44 35 pm 2

screen shot 2019-02-22 at 9 45 05 pm 2

### Alternate Designs

The ora spinner could be used to its full potential here but it would require what looks like a complete rewrite of the CLI handlers. Ora may be able to takeover the entire CLI with its own success check marks, failure marks, spinner, and text. It seems like that would be unnecessary work to me. I'm happy to look into it deeper if that's desired though.

### Benefits

It frankly looks exactly the same as it did using the elegant-spinner package. Using ora would allow you to easily change the spinner color with this implementation though. Maybe a green spinner would better match the Neo-One colors? These are the available colors: black, red, green, yellow, blue, magenta, cyan, white, gray.

### Possible Drawbacks

According to https://bundlephobia.com the ora package is 35.7kb minified, with 6 dependencies, whereas the elegant-spinner package is 343b minified, with 0 dependencies. elegant-spinner is extremely simple to implement whereas ora may take a bit to sort through if this ever needs to be updated somehow.

### Applicable Issues

No issues that I can see. I ran `yarn remove elegant-spinner` which changed the `yarn.lock` file though. I suppose that _shouldn't_ cause any issues but I put that change in a separate commit so it can be easily examined.


danwbyrne:
Hi @spencercorwin this is something I looked at doing when I had to make the @types for elegant-spinner. Your last point about the bundle differences is *somewhat* concerning; we are on a bit of a crusade at the moment to reduce our bundle size/dependency list.

When I have some time I'll take a look at this PR :) 👍

spencercorwin:
@danwbyrne For sure. I'm around, so let me know how I can help. It won't hurt my feelings if this PR ends up being useless and rejected.

PR #1136, opened 2019-02-27
refactor(gulp browserify): browserify now runs as part of the build command

afragapane:
### Description of the Change
browserify now runs as part of the build command and will publish properly.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -95,9 +95,27 @@ const createCustomTransform = (context) => {
   };
 };
 
-const createFormatDistName = ({ target, module }) => `neo-one-${target}-${module}`;
+const createFormatDistName = ({ main, target, module, browser }) => {
+  if (main) {
+    return 'neo-one';
+  }
+  if (browser) {
+    return 'neo-one-browserify';
+  }
 
-const createFormatName = ({ target, module }) => `${target}-${module}`;
+  return `neo-one-${target}-${module}`;
+};
+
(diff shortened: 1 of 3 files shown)


gulpfile.js
```diff
@@ -632,97 +666,85 @@ const copyBrowserifyTypesFiles = (inputPath, outputPath) => {
...
+                      }),
+                    );
+                    reject(new Error('Webpack bundling failed.'));
+                  } else {
+                    resolve(stats);
+                  }
+                },
+              ),
+            );
+          } catch (error) {
```
dicarlo2: Remove this try/catch

gulpfile.js
```diff
@@ -632,97 +666,85 @@ const copyBrowserifyTypesFiles = (inputPath, outputPath) => {
...
+const copyBrowserifyTypes = ((cache) =>
+  memoizeTask(cache, async function copyBrowserifyTypes(format) {
+    if (format.browser) {
+      const inputPath = path.resolve(DIST, 'neo-one', 'packages');
+      const outputPath = getDest(format);
+
+      const allPackages = await fs.readdir(inputPath);
+      await Promise.all(
+        allPackages.map(async (pkgName) => {
+          if (CLIENT_FULL_PACKAGES.includes(getPkgName(pkgName))) {
```
dicarlo2: We should generally be using a set for O(1) access rather than O(n).

PR #1167, opened 2019-03-04
fix(website): fix overflow styling

danwbyrne:
fix overflow styling in website, also update some deps in the lockfile + add a resolution for react to root package.json.


davemneo:
LGTM

PR #1177, opened 2019-03-05
doc(setup): include code-gen config

davemneo:
Update ` tutorial.md ` to include details about code-gen configuration

Testing, many of these you just need to run one time on a fresh clone
``` yarn
yarn build
yarn website:start-workers
yarn website:start-tools
yarn website:start-static
yarn website:start-server
yarn website:start-workers
```
navigate browser to:
` localhost:3000 `
be sure to clear application cache

New doc's page looks like this:
image

Tutorial looks like:
image


davemneo:
Took @afragapane's recommendation to add a Docs page including all current configuration options.

packages/neo-one-website/docs/2-advanced-guides/10-configuration-options
```diff
@@ -0,0 +1,41 @@
...
+
+Configure NEO•ONE for your project.
+
+## Config File
+
+An [initialized environment](https://neo-one.io/tutorial#Setup-for-the-Tutorial) will have a ` .onerc ` file within the project root.
+
+### `codegen`
+
+Keys with `codegen` allow you to configure which `language`, front-end `framework` and whether to generate code for the `browser`.
```
afragapane: I suggest changing this whole block a bit . Something like: " 
There are three options within `codegen` which define options for the code generated by NEO•ONE on compilation.  
*`language`: `typescript` | `javascript`.  Determines whether generated code will be in `typescript` or `javascript`.  Defaults to `typescript`.
*`framework`: `none` | `react` | `angular` | `vue`.  Determines which front-end framework to generate helper components for.  Defaults to `none`.  For more information on what is generated for each of these frameworks, see our docs on [react](https://neo-one.io/docs#react), [angular](https://neo-one.io/docs#angular), [vue](https://neo-one.io/docs#vue).
*`browser`: `boolean`.  Setting `browser` to `true` generates browser-compatible code by referencing the pre-bundled `@neo-one/client-browserify` instead of `@neo-one/client` in the generated files.  Defaults to `false`.  Note, this option is intended for use with react-native based frameworks such as [Expo](https://expo.io) which do not allow nodejs builtin modules. 
"

packages/neo-one-website/docs/2-advanced-guides/10-configuration-options
```diff
@@ -0,0 +1,41 @@
...
+* `language` may be  `typescript` or `javascript`
+* `framework` currently supports choosing one of `none`, `react`, `vue` and `angular`
+* `browser` is a boolean; either `true` or `false`
+
+### `paths`
```
afragapane: This should be changed to match the style of `codegen` . Something like: "
There are two options within `path` which defined where NEO•ONE will look for smart contracts to compile and where it will output generated code.  
*`contracts`: `string`.  Path at which NEO•ONE will search for smart contracts to compile. Defaults to `one/contracts`.
*`generated`: `string`. Path at which NEO•ONE will output generated code.  Defaults to `one/generated`.
"

packages/neo-one-website/docs/2-advanced-guides/10-configuration-options
```diff
@@ -0,0 +1,41 @@
...
+By default NEO•ONE compiles from source files found in `one/contracts` and writes the output to `one/generated`.  This configuration may be changed by updating the `contracts` and `generated` keys under `paths`.
+
+### `browser`
```
afragapane: Remove, this key is in codegen.

packages/neo-one-website/docs/2-advanced-guides/10-configuration-options
```diff
@@ -0,0 +1,41 @@
+---
+slug: config options
```
afragapane: I think to match all the other slugs, this should be `config-options`

packages/neo-one-website/tutorial/tutorial.md
```diff
@@ -68,6 +68,8 @@ npm install @neo-one/cli @neo-one/client @neo-one/smart-contract @neo-one/smart-
 
 We recommend taking a moment to [setup your editor](/docs/environment-setup#Editor-Setup) to take advantage of inline NEO•ONE compiler diagnostics.
 
+5. Review the available [configuration options](/docs/config%20options) and make the necessary modifications to the `.onerc` file.
```
afragapane: Maybe note that they may want to change `framework` to `react` if they want to follow along with the tutorial since it leverages react and the default is `none`.

afragapane: Also see above note about the slug. `config-options`

packages/neo-one-website/docs/2-advanced-guides/10-configuration-options
davemneo: very helpful, thank you!

packages/neo-one-website/docs/2-advanced-guides/10-configuration-options
```diff
@@ -0,0 +1,41 @@
...
+
+### Example NEO•ONE configuration file
+
+The `.onerc` configuration file could look something like this:
+```json
+{
+  "codegen": {
+    "language": "typescript",
+    "framework": "react",
+    "browser":  true
```
afragapane: Change this to false.  The default is false and they really shouldn't be using the browser option except for specific use cases.  So for people copy-pasting from here or something, we would want it false.

packages/neo-one-website/docs/2-advanced-guides/10-configuration-options
```diff
@@ -0,0 +1,41 @@
...
+  },
+}
+```
+
+### `codegen`
+
+Keys with `codegen` allow you to configure which `language`, front-end `framework` and whether to compile with `browser` compatibility.
+* `language`: `string` =  `typescript` | `javascript`
+* `framework`: `string` =  `none` | `react` | `vue` | `angular`
+* `browser`:  `boolean` = `true` | `false`
```
afragapane: While the language and framework may be self explanatory, the browser key needs to be explained in more detail.  We don't want people to use this option unless they absolutely need to.  Use what I wrote in my previous comment under the browser key.

afragapane: I also think we should specify the defaults somehow.

packages/neo-one-website/docs/2-advanced-guides/10-configuration-options
```diff
@@ -0,0 +1,41 @@
...
+
+### `paths`
+There are two options within `path` which defined where NEO•ONE will look for smart contracts to compile and where it will output generated code.
```
afragapane: defined -> define

packages/neo-one-website/tutorial/tutorial.md
```diff
@@ -68,6 +68,9 @@ npm install @neo-one/cli @neo-one/client @neo-one/smart-contract @neo-one/smart-
 
 We recommend taking a moment to [setup your editor](/docs/environment-setup#Editor-Setup) to take advantage of inline NEO•ONE compiler diagnostics.
 
+5. Review the available [configuration options](/docs/config-options) and make the necessary modifications to the `.onerc` file.
```
afragapane: What necessary modifications?  Maybe instead of the react specific note below, you just say something like "Here are the config options used in this tutorial." and then give the example config with the correct options?

Take or leave this comment, just think maybe we could be a little clearer here.

davemneo: How's this... 
https://github.com/neo-one-suite/neo-one/pull/1177/files#diff-441447123cbc184d48bd6602a2c76df1R72

packages/neo-one-website/docs/2-advanced-guides/10-configuration-options
davemneo: added defaults:
https://github.com/neo-one-suite/neo-one/pull/1177/files#diff-172ea3c052d9b136cba8549a61c6fd9dR33

packages/neo-one-website/docs/2-advanced-guides/10-configuration-options
davemneo: updated: https://github.com/neo-one-suite/neo-one/pull/1177/files#diff-172ea3c052d9b136cba8549a61c6fd9dR21

PR #1178, opened 2019-03-05
fix(build): fix developerTools building

danwbyrne:
### Description of the Change

`@neo-one/developer-tools` is built in a strange way compared to other packages. Since we want to have it both minimized and usable as `.ts` for production I had to separate a few things out in `./gulpfile.js`.
- `compileDeveloperTools` is now 2 parts, `compileDeveloperToolsDeclarations` and `compileDeveloperToolsBundle`.

`compileDeveloperToolsDeclarations` just builds the declarations and `compileDeveloperToolsBundle` just builds the bundled `index.js`. The `package.json`, `README.md`, etc get moved over the same as before.

### Test Plan

checkout master and `yarn build`, inspect the `developer-tools` package and note how *jank* it is.

now checkout this and `yarn build`, see how crisp and clean it is.

### Alternate Designs

The way the build is done could be different I suppose.

### Benefits

Developer Tools is still minimized and *typed*, as expected.

### Possible Drawbacks

I need to test it in prod.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -246,7 +246,7 @@ const globs = {
     `!packages/*/src/__tests__/**/*`,
     `!packages/*/src/__e2e__/**/*`,
     `!packages/*/src/bin/**/*`,
-    `!packages/neo-one-developer-tools/src/*.ts`,
+    `!packages/neo-one-developer-tools/src/*.{ts,tsx}`,
     `!packages/neo-one-developer-tools-frame/src/*.ts`,
     `!packages/neo-one-smart-contract-lib/src/*.ts`,
     `!packages/neo-one-server-plugin-wallet/src/contracts/*.ts`,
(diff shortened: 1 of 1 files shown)


davemneo:
LGTM

gulpfile.js
```diff
@@ -493,9 +512,17 @@ const buildAllNoDeveloperTools = ((cache) =>
     )(done);
   }))({});
 
+const compileDeveloperTools = ((cache) =>
+  memoizeTask(cache, function compileDeveloperTools(format, done, type) {
+    return gulp.parallel(
+      compileDeveloperToolDeclarations(format, type),
+      compileDeveloperToolsBundle(format, type),
```
afragapane: `type` is passed to `compileDeveloperToolsBundle`, but `compileDeveloperToolsBundle` does not require a `type` argument.

Issue #1175, opened 2019-03-05
doc(setup): add config for 'one' 'codegen' 


davemneo:
### Description
Document the how the neo-one code generator picks up configuration.

PR #1180, opened 2019-03-06
feat(config): add default .onerc config

davemneo:
write ` .onerc ` with default config values.

### Requirements

Produce a .onerc config file with when ` yarn neo-one init ` is called

### Description of the Change

adds a new ` .onerc ` config file during the init process containing the default configuration as specified in ` packages/neo-one-server-plugin-project/src/types.ts `

### Test Plan

    yarn neo-one init

see that the ` .onerc ` file contains the appropriate defaults as defined by ` packages/neo-one-server-plugin-project/src/types.ts `

fixes: #1175

### DIFF
--- a/packages/neo-one-server-plugin-project/src/initCommand.ts
+++ b/packages/neo-one-server-plugin-project/src/initCommand.ts
@@ -7,37 +7,55 @@ import jsonFile from 'json-file-plus';
 import * as path from 'path';
 import { loadProjectConfig } from './utils';
 
+interface InitDataFileOpts {
+  readonly folder: string;
+  readonly name: string;
+  // tslint:disable-next-line:no-any
+  readonly data: string;
+  readonly cli: InteractiveCLIArgs['cli'];
+}
+export const initDataFile = async ({ folder, name, data, cli }: InitDataFileOpts) => {
+  const filePath = path.resolve(folder, name);
+  await fs.ensureDir(folder);
+  const createdUpdated = fs.existsSync(filePath) ? 'updated' : 'created';
+  await fs.writeFile(
+    filePath,
+    JSON.stringify(
+      JSON.parse(data),
+      // tslint:disable-next-line:no-null-keyword
+      null,
+      2,
+    ),
+  );
+  cli.print(`${createdUpdated} ${name} at ${filePath}.`);
+};
+
 export const initCommand = ({ cli }: InteractiveCLIArgs) =>
   cli.vorpal.command('init', 'Initialize neo-one in the current directory.').action(async () => {
     const cwd = process.cwd();
     const projectConfig = await loadProjectConfig(cwd);
     const tsconfigPath = path.resolve(projectConfig.paths.contracts, 'tsconfig.json');
     const contractsPath = path.dirname(tsconfigPath);
-
-    await fs.ensureDir(contractsPath);
-    await fs.writeFile(
-      tsconfigPath,
-      JSON.stringify(
-        {
-          compilerOptions: {
-            ...COMPILER_OPTIONS,
-            target: 'esnext',
-            module: 'esnext',
-            moduleResolution: 'node',
-            jsx: 'react',
-            plugins: [
-              {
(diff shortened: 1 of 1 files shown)


davemneo:
what's the "no-any" rule? (see lint result: https://circleci.com/gh/neo-one-suite/neo-one/36117?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link)

packages/neo-one-server-plugin-project/src/initCommand.ts
```diff
@@ -7,37 +7,54 @@ import jsonFile from 'json-file-plus';
...
-        // tslint:disable-next-line:no-null-keyword
-        null,
-        2,
-      ),
-    );
-    cli.print(`Created tsconfig.json at ${tsconfigPath}.`);
+        ],
+      },
+    };
+    const rcData = { codegen: projectConfig.codegen };
```
afragapane: I think we actually want the .neorc file to contain the whole projectConfig, not just the codegen piece.

packages/neo-one-server-plugin-project/src/initCommand.ts
```diff
@@ -7,37 +7,54 @@ import jsonFile from 'json-file-plus';
 import * as path from 'path';
 import { loadProjectConfig } from './utils';
 
+interface InitDataFileOpts {
+  readonly folder: string;
+  readonly name: string;
+  // tslint:disable-next-line:no-any
+  readonly data: any;
```
afragapane: We can maybe make this more strict than `any` since it is either expecting a ProjectConfig https://github.com/neo-one-suite/neo-one/blob/6030f892c6d77419cda83bd0bcb55e896118d0e1/packages/neo-one-server-plugin-project/src/types.ts#L36 or a tsconfig.

packages/neo-one-server-plugin-project/src/initCommand.ts
afragapane: I guess it's not necessary, but imo might as well be explicit since we have so few options.  Easier for them to figure out what config options they may want to change without digging through documentation if we just print the whole thing from the start.

packages/neo-one-server-plugin-project/src/initCommand.ts
davemneo: we could do the ProjectConfig, but TSConfig as a type does not appear to exist, and I'd ask @dicarlo2  if he thinks it's worth while to define the shape of a TSConfig.

dicarlo2: Just make `data` a string and `JSON.stringify` before passing it into the function.

packages/neo-one-server-plugin-project/src/initCommand.ts
dicarlo2: I'm not sure I quite understand what this is doing - we load the project config from the cwd and then write it back out?

dicarlo2:
cc @danwbyrne @afragapane to take a look at my comments.

packages/neo-one-server-plugin-project/src/initCommand.ts
davemneo: Pulling in defaults.  If something is defined in the existing `.onerc`, it is used, otherwise use defaults.

davemneo: If this is Ideal, I can see where to get the default paths, otherwise "default paths" are not part of the schema definition.

afragapane: The default paths will already been included in the `projectConfig`.  The `projectConfig` is loaded by `loadProjectConfig` which will set default paths if they are not defined.  So, if you want to make this chance, all you need to do is pass the whole `projectConfig` instead of extracting out the codegen piece.

packages/neo-one-server-plugin-project/src/initCommand.ts
davemneo: +JSON.stringify(data) 
+JSON.parse(data)

PR #1202, opened 2019-03-13
fix(build): fix developerTools building and the node image build

danwbyrne:
Moved some more things around and add a filter to stop `.js` files from getting generated in `DeveloperTools` since we just want the single `index.js` generated in a bundle. Also had to add some files to the Dockerfile since `DeveloperTools` is exported from client now. 

Docker container should build again now without affecting our build.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -246,7 +246,6 @@ const globs = {
     `!packages/*/src/__tests__/**/*`,
     `!packages/*/src/__e2e__/**/*`,
     `!packages/*/src/bin/**/*`,
-    `!packages/neo-one-developer-tools/src/*.ts`,
     `!packages/neo-one-developer-tools-frame/src/*.ts`,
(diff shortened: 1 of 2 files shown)


danwbyrne:
bors+

danwbyrne:
> One question, what is the purpose of the name changes?

Because of the changes we made the names signify different things now. Before we excluded all of `DeveloperTools` from the node-build because we didn't need it and none of the required packages for the node use it. Now that `@neo-one/client` exports `DeveloperTools` we need to build *just* its declaration files for the node, however for the full build we also generated the `DeveloperToolsBundle` which relies on the minimized file gen'd from `compileDeveloperToolsFrame` (which we don't run in the node).

TL;DR after the refactor only the `DeveloperToolsBundle` isn't generated in the node-build, so I've relabeled the functions appropriately.

PR #1208, opened 2019-03-14
fix(jest): fix a flaky crypto test - createKeyPair

danwbyrne:
### Description of the Change

`crypto.createKeyPair` used to fail occasionally. `ec().genKeyPair()` can sometimes return a bad value, so we just need to `validate()` it before we attempt to use it. This fixes the issue.

Also, I saw an e2e test fail sometime recently I think with the EADDRINUSE again, so going to pump the range on `getAvailableMinPortForRange` and see if that nips it in the bud for good.

### Test Plan

I tested it by isolating the currently failing test and running it x *N* to watch it fail. Then I found a solution in the `validate` function.

### Alternate Designs

??

### Benefits

FLAKY TESTS ARE SCARED OF ME

### Possible Drawbacks

PRs might go through faster now, not sure if we can keep up with them in terms of reviews. ;)

### DIFF
--- a/packages/neo-one-client-common/src/crypto.ts
+++ b/packages/neo-one-client-common/src/crypto.ts
@@ -22,11 +22,10 @@ import { p256 } from './precomputed';
 import { ScriptBuilder } from './ScriptBuilder';
 
 // tslint:disable-next-line no-let
-let ecCache: any;
+let ecCache: EC | undefined;
 const ec = () => {
   if (ecCache === undefined) {
-    // tslint:disable-next-line no-any
-    ecCache = new EC(p256) as any;
+    ecCache = new EC(p256);
   }
 
   return ecCache;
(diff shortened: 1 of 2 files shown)


packages/neo-one-client-common/src/crypto.ts
```diff
@@ -22,11 +22,10 @@ import { p256 } from './precomputed';
 import { ScriptBuilder } from './ScriptBuilder';
 
 // tslint:disable-next-line no-let
-let ecCache: any;
+let ecCache: EC | undefined;
 const ec = () => {
   if (ecCache === undefined) {
-    // tslint:disable-next-line no-any
-    ecCache = new EC(p256) as any;
+    ecCache = new EC(p256);
```
davemneo: Could this be simplified by assigning ` new EC(p256) ` as the default ` ecCache `:

```typescript
let ecCache: EC = new EC(p256);
const ec = () => ecCache;
```

danwbyrne: no, we specifically do it this way so that `ec` is a singleton not created until something needs it. The changes I made here were only because EC is typed now and we don't need the `any`.

davemneo: LGTM

davemneo:
LGTM

PR #1211, opened 2019-03-15
fix(codecov): put 'checkports' inside scripts/e2e

danwbyrne:
title, codecov was tracking it now it won't.


davemneo:
FWIW, LGTM

Issue #292, opened 2018-07-02
Add support for mnemonic phrase keystore


dicarlo2:
Some interesting reads from Bitcoin Improvement Propositions:

https://github.com/bitcoin/bips/blob/master/bip-0032.mediawiki
https://github.com/bitcoin/bips/blob/master/bip-0039.mediawiki

BIP32 seems extensive so I have quite a bit of reading to do there but it seems related?
BIP39 is details about mnemonic implementations with some additional libraries.

https://github.com/trezor/python-mnemonic here is an actual implementation of it in python.

danwbyrne:
Some interesting reads from Bitcoin Improvement Propositions:

https://github.com/bitcoin/bips/blob/master/bip-0032.mediawiki
https://github.com/bitcoin/bips/blob/master/bip-0039.mediawiki

BIP32 seems extensive so I have quite a bit of reading to do there but it seems related?
BIP39 is details about mnemonic implementations with some additional libraries.

danwbyrne:
Add `addAccount` to the keystore and to localUserAccountProvider similar to how `updateAccountName` is implemented now.

PR #1247, opened 2019-03-26
feat(client): Add HDKeyStore & refactor LedgerKeyStore

danwbyrne:
### Description of the Change

BIG.

Adds the `HDStore`, `HDHandler` and `HDKeyStore` class as well as the crypto functionality to handle it. `LocalHDStore` / `LocalHDHandler` are our implementations of the BIP32 spec: https://github.com/bitcoin/bips/blob/master/bip-0032.mediawiki

**Breaking**: refactors `ledgerKeyStore` into an instance of `HDKeyStore`.

### Test Plan

multiple tests are available, although I still need to go through with wallaby and make sure I'm fully covering everything. Can run `yarn test Handler` or `yarn test Store`.

### Alternate Designs

:thinking:

### Benefits

Adds support for HD Wallets.

### Applicable Issues

#292


packages/neo-one-client-core/src/user/keystore/LocalHDStore.ts
```diff
@@ -0,0 +1,564 @@
...
+  readonly resolve: () => void;
+  readonly reject: (error: Error) => void;
+}
+
+export interface NodeAccount {
+  readonly node: HDNode;
+  readonly path: LocalPath;
+}
+
+export type NodeAccounts = { [Index in string]: NodeAccount };
```
dicarlo2: nit: `{ [index: string]: NodeAccount}`, or even `Record<string, NodeAccount>`

dicarlo2: here and elsewhere

packages/neo-one-client-common/src/crypto.ts
```diff
@@ -575,6 +586,192 @@ const isSignatureContract = (script: Buffer) => script.length === 35 && script[0
 
 const isStandardContract = (script: Buffer) => isSignatureContract(script) || isMultiSigContract(script);
 
+const HARDENED_KEY_OFFSET = 0x80000000;
+const EXTENDED_KEY_BYTES = 78;
+
+// Need eyes, according to https://github.com/bitcoin/bips/blob/master/bip-0043.mediawiki#node-serialization this isn't special.
```
dicarlo2: Seems legit, go ahead and leave the comment with the url though so we know why this is here later.

packages/neo-one-client-common/src/crypto.ts
```diff
@@ -575,6 +586,192 @@ const isSignatureContract = (script: Buffer) => script.length === 35 && script[0
 
 const isStandardContract = (script: Buffer) => isSignatureContract(script) || isMultiSigContract(script);
 
+const HARDENED_KEY_OFFSET = 0x80000000;
```
dicarlo2: Can you add references to the documentation for all of the constants in this PR (similar to below).

packages/neo-one-client-common/src/crypto.ts
```diff
@@ -575,6 +586,192 @@ const isSignatureContract = (script: Buffer) => script.length === 35 && script[0
...
+  readonly parentFingerprint?: Buffer;
+  readonly index: number;
+  readonly chainCode: Buffer;
+}
+
+const getFingerprint = (value: Buffer) => hash160(value).slice(0, 4);
+
+const parseMasterSeed = (seedIn: Buffer | string): HDNode => {
+  const seed = seedIn instanceof Buffer ? seedIn : Buffer.from(seedIn, 'hex');
+  const hmac = hmacSha512('NEO Seed', seed);
```
dicarlo2: Why `'NEO Seed'`?

packages/neo-one-client-core/src/__ledger_tests__/NanoLedgerStore.test.ts
```diff
@@ -1,15 +1,15 @@
 import { publicKeyToAddress } from '@neo-one/client-common';
-import { LedgerHandler } from '../user/keystore/LedgerHandler';
+import { NanoLedgerStore } from '../user/keystore/NanoLedgerStore';
```
dicarlo2: `LedgerNano`? 😃

packages/neo-one-client-core/src/user/keystore/HDKeyStore.ts
```diff
@@ -0,0 +1,224 @@
...
+  readonly [Network in string]?: { readonly [Address in string]?: HDAccount<Identifier> }
+};
+
+export interface HDProvider {
+  readonly getAccount: (network: NetworkType, address: AddressString, monitor?: Monitor) => Promise<Account>;
+  readonly getNetworks: () => ReadonlyArray<NetworkType>;
+  readonly networks$: Observable<ReadonlyArray<NetworkType>>;
+}
+
+export interface HDStore<Identifier> {
```
dicarlo2: Is this interface necessary? Seems like we should be using `HDHandler` everywhere instead?

dicarlo2: Hmm, maybe this is just a common type you're using? Move it into a `types.ts` file since it's not used anywhere in this file.

packages/neo-one-client-core/src/user/keystore/LedgerKeyStore.ts
```diff
@@ -1,306 +1,12 @@
...
-import { BehaviorSubject, Observable, Subscription } from 'rxjs';
-import { distinctUntilChanged, map } from 'rxjs/operators';
-import { UnknownAccountError, UnknownNetworkError } from '../../errors';
-import { KeyStore } from '../LocalUserAccountProvider';
+import { HDKeyStore, HDProvider } from './HDKeyStore';
 import { LedgerHandler } from './LedgerHandler';
+import { NanoLedgerStore } from './NanoLedgerStore';
 
-type Ledgers = { readonly [Network in string]?: { readonly [Address in string]?: Ledger } };
+export const LedgerKeyStore = (provider: HDProvider) => {
```
dicarlo2: nit: `createLedgerKeyStore`

packages/neo-one-client-core/src/__ledger_tests__/NanoLedgerStore.test.ts
danwbyrne: WOW YOU ARE GONNA COME AT ME WITH THIS

packages/neo-one-client-common/src/crypto.ts
danwbyrne: The seed name doesn't matter. I only copied this from 'BTC Seed' and changed it to `NEO`.

danwbyrne:
> High level question: how confident are you in the implementation? Are we prepared to manage millions in funds with this? Should we get this audited externally before shipping? Are there a standardized set of tests we can verify the implementation against?

There weren't any set like we had with the Ledger for example, no. As far as confidence? I'm like 85%? I definitely think before using it in Neotracker maybe have it audited? Or at the least have someone make a big E2E test. I don't see why we couldn't release it now with a `beta` tag in the documentation or equivalent.

PR #1248, opened 2019-03-26
fix(website): add dependency updates and general fixes

danwbyrne:
### Description of the Change

Brings everything up to date for deploying the website on dependencies which were updated underneath us; Overall developing the website and deploying it should be easier and more consistent between dev/prod now (like it was before).

### Benefits

I can deploy the website now :) we still need to publish though since changes affected more than just the `@neo-one/website` package.


packages/neo-one-developer-tools-frame/src/Toasts.tsx
```diff
@@ -19,6 +19,9 @@ interface Props {
 }
 
 export function Toasts({ resizeHandler }: Props) {
+  // tslint:disable-next-line:no-unused
+  const [toasts, _addToast, removeToast] = useToasts();
```
davemneo: Reviewed the _unused var patten:
// ok

packages/neo-one-website/courses/0-tokenomics/2-launch/1-balance/src/index.tsx
```diff
@@ -8,11 +8,13 @@ import { ContractsProvider } from '../one/generated';
 import { ICO } from './ICO';
 
 const App = (
+  // <StyleSheetManager target={document.head}>
```
davemneo: Why?

packages/neo-one-website/src/components/reference/components/ParameterPropertyList.tsx
```diff
@@ -17,12 +17,12 @@ const ParameterLayout = styled(Box)`
 `;
 
 interface Props {
-  readonly values: ReadonlyArray<Parameter | Property>;
+  readonly values?: ReadonlyArray<Parameter | Property>;
```
davemneo: Why is this optional if you're assigning a default value to it anyway?

danwbyrne: Its optional just to match the usage of the input which is optional (i'd have to track it down but if you peek references its called like `functiondata.<something>`).

Issue #1254, opened 2019-03-27
Allow conversion between Arrays & Struct stack items


afragapane:
### Description
neo-one uses an OPCode `PACKSTRUCT` which does not exist in the neo-vm.  We attempted to add this OPCode to the neo-vm (https://github.com/neo-project/neo-vm/issues/51, https://github.com/neo-project/neo-vm/pull/88).  It was suggested that instead of using `PACKSTRUCT`, to instead use a combination of `PACK` + `NEWSTRUCT` and modify `NEWSTRUCT` to allow conversion from an array.

PR #1256, opened 2019-03-27
refactor(PACK+NEWSTRUCT): Replace PACKSTRUCT with PACK+NEWSTRUCT

afragapane:
### Description of the Change
Removes the PACKSTRUCT Opcode and modifies NEWSTRUCT/NEWARRAY to be able to convert an existing array into a struct and struct into array.  This allows the use of a PACK+NEWSTRUCT pair of opcodes in place of PACKSTRUCT.

### Test Plan
Added tests for the conversions in NEWSTRUCT and NEWARRAY.  Updated existing tests to reflect the slightly higher GAS costs of neo-one compiled contracts.

### Alternate Designs
Previously the plan was to implement our PACKSTRUCT in the neo-vm as per https://github.com/neo-project/neo-vm/issues/51 and https://github.com/neo-project/neo-vm/pull/88.  This new solution was suggested and implemented here: https://github.com/neo-project/neo-vm/pull/91.
### Possible Drawbacks
Very slightly higher GAS cost for neo-one compiled smart contracts.  

### Applicable Issues
https://github.com/neo-one-suite/neo-one/issues/1254

### DIFF
--- a/packages/neo-one-node-vm/src/opcodes.ts
+++ b/packages/neo-one-node-vm/src/opcodes.ts
@@ -473,6 +473,31 @@ const functionCallIsolated = ({ name }: { readonly name: OpCode }): OpCreate => 
   },
 });
 
+const newArrayOrStruct = ({ name }: { readonly name: 'NEWARRAY' | 'NEWSTRUCT' }) =>
+  createOp({
+    name,
+    in: 1,
+    out: 1,
+    invoke: ({ context, args }) => {
+      let results;
+      if (args[0].isArray()) {
+        const array = args[0].asArray();
+        results = name === 'NEWARRAY' ? [new ArrayStackItem(array)] : [new StructStackItem(array)];
+      } else {
+        const count = vmUtils.toNumber(context, args[0].asBigInteger());
+
+        if (count > MAX_ARRAY_SIZE) {
+          throw new ContainerTooLargeError(context);
+        }
+
+        const fill = _.range(0, count).map(() => new BooleanStackItem(false));
+        results = name === 'NEWARRAY' ? [new ArrayStackItem(fill)] : [new StructStackItem(fill)];
+      }
+
+      return { context, results };
+    },
+  });
+
 const JMP = jump({ name: 'JMP' });
 
 const OPCODE_PAIRS = ([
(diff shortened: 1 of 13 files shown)


davemneo:
LGTM

packages/neo-one-node-vm/src/opcodes.ts
```diff
@@ -473,6 +473,31 @@ const functionCallIsolated = ({ name }: { readonly name: OpCode }): OpCreate =>
   },
 });
 
+const newArrayOrStruct = ({ name }: { readonly name: 'NEWARRAY' | 'NEWSTRUCT' }) =>
```
danwbyrne: Curious why this needs to be deconstructed like this? Would:
```
const newArrayOrStruct = (name: 'NEWARRAY' | 'NEWSTRUCT') => ...
```
not work? Maybe its typing issue.

packages/neo-one-server-plugin-project/src/__data__/icojs/one/generated/sourceMaps.js
```diff
@@ -1,15 +1,15 @@
-/* @hash 6e6420d58bfda29c46d8be62f51891e8 */
```
danwbyrne: I've been doing `git checkout` on these changes when they come up. Its not a big deal just extra stuff on the PR.

packages/neo-one-node-vm/src/opcodes.ts
```diff
@@ -473,6 +473,31 @@ const functionCallIsolated = ({ name }: { readonly name: OpCode }): OpCreate =>
...
+
+        const fill = _.range(0, count).map(() => new BooleanStackItem(false));
+        results = name === 'NEWARRAY' ? [new ArrayStackItem(fill)] : [new StructStackItem(fill)];
+      }
+
+      return { context, results };
+    },
+  });
+
 const JMP = jump({ name: 'JMP' });
```
danwbyrne: I guess my below complaint is mute if its just the standard pattern for these?

afragapane: *shrugs*, yeah I was just matching the style of other opcodes.

packages/neo-one-server-plugin-project/src/__data__/icojs/one/generated/sourceMaps.js
afragapane: Good call, I'll do that and update.

PR #1304, opened 2019-04-10
feat(mixin-secondary): add Secondary

davemneo:
This class provides a means to mark an address as the primary caller of this contract.  Extending this class and adding `this.primaryOnly();` to the beginning of all public functions will throw an error anytime an address other than the `primary` makes requests.

For example, this could be used to allow a CrowdsaleContract of a given address to be the primary caller for a Token, and the Token only allows the CrowdsaleContract or Owner to execute the Mint function.

Replaces DesignatedCaller with functionality appropriate for the OpenZeppelin template set.


packages/neo-one-smart-contract-lib/src/ownership/Secondary.ts
```diff
@@ -0,0 +1,33 @@
+import { Address, constant, createEventNotifier, SmartContract } from '@neo-one/smart-contract';
+
+/**
```
dicarlo2: nit: remove the comment (since we don't parse and show these anywhere)

packages/neo-one-smart-contract-lib/src/ownership/Secondary.ts
```diff
@@ -0,0 +1,33 @@
...
+ * @title Designated Caller
+ * @dev This contract can only be called by 1 other address.
+ */
+const primaryTransferred = createEventNotifier<Address, Address>('primary_transferred', 'from', 'recipient');
+
+export class Secondary extends SmartContract {
+  protected mutablePrimary: Address;
```
dicarlo2: This should probably use the same scheme as `Ownable` in terms of setting the initial primary. Also, any reason why this is called `mutablePrimary` (and `primary` used throughout) rather than `secondary`?

davemneo: @dicarlo2 

RE: Secondary vs Primary: This contract *is* secondary to the Primary contract.  This follows the model here:  https://github.com/OpenZeppelin/openzeppelin-solidity/blob/master/contracts/ownership/Secondary.sol

davemneo: Updated initialPrimary scheme to be like Ownable.

davemneo:
Unless there are any additional changes, all review requests have been addressed.

PR #1057, opened 2019-02-07
feat(contract-mixins): add TransferableContract

davemneo:
Branched off https://github.com/neo-one-suite/neo-one/pull/1054 NEP5 Token Rename 

### Description of the Change
Introduce a SmartContract Mix-in that provides basic "transferable contract" functionality with a suite of tests.  

### Test Plan
Run the included test suite:

     yarn jest packages/neo-one-smart-contract-lib/src/__tests__/TestTransferableOwnership.test.ts

PR #1305, opened 2019-04-10
feat(mixin-ownable): add Ownable mixin and tests

davemneo:
# Overview 
This class provides a means to assign an address as the owner of a contract. Extending this class and adding ` this.ownerOnly(); ` to the beginning of all public functions will throw an error anytime an address other than the primary makes requests.  

### Public
`owner()` - returns the current owner address or undefined.
##### Manage contract Ownership with...
 `renounceOwnership()` - Allows only the owner to contract to become ownerless, this will cause all `ownerOnly()` calls to fail, this is not reversible.  Note: the `initialOwner` will maintain still contain the initial address.  
 `transferOwnership()` - Only the owner may transfer ownership to a new 

### Internal
`ownerOrThrow()` - returns the current owner address or throws an error.
`onlyOwner()` - if the owner address did not call this contract an error is thrown.

### Published Events:
`tokensPurchased` with purchaser, beneficiary and amount of tokens bought.

## Configuration:
Be sure that any contract extending this `Ownable` mix-in sets the `initialOwner` thusly: 
``` typescript:
export class MyContract extends Ownable(SmartContract) {
  public constructor(protected readonly initialOwner = Deploy.senderAddress) {
    super();
  }
  // .... awesome smart contract code ... 
}
```

## Tests

    yarn jest TestOwnable


packages/neo-one-smart-contract-lib/src/ownership/Ownable.ts
```diff
@@ -0,0 +1,52 @@
+import { Address, constant, createEventNotifier, SmartContract } from '@neo-one/smart-contract';
+
+/**
+ * @title Ownable
+ * @dev A ownable contract has an owner address
+ */
+
+/* tslint:disable-next-line: variable-name */
+const transferred_ownership = createEventNotifier<Address | undefined, Address>('transferred ownership', 'from', 'to');
```
dicarlo2: This should be something like

`const transferOwner = createEventNotifier<Address | undefined, Address>('transfer_owner', 'from', 'to');`

The variable name should follow the same rules as any other variable (camelcase) and the event should always be snake case.

packages/neo-one-smart-contract-lib/src/ownership/Ownable.ts
```diff
@@ -0,0 +1,52 @@
+import { Address, constant, createEventNotifier, SmartContract } from '@neo-one/smart-contract';
+
+/**
```
dicarlo2: Remove, this doesn't do anything.

packages/neo-one-smart-contract-lib/src/ownership/Ownable.ts
```diff
@@ -0,0 +1,52 @@
...
+ * @dev A ownable contract has an owner address
+ */
+
+/* tslint:disable-next-line: variable-name */
+const transferred_ownership = createEventNotifier<Address | undefined, Address>('transferred ownership', 'from', 'to');
+
+export function Ownable<TBase extends Constructor<SmartContract>>(base: TBase) {
+  class OwnableClass extends base {
+    protected mutableOwner: Address | undefined;
+    public constructor(...args: any[]) {
```
dicarlo2: Mixins shouldn't use constructor arguments because there's no guarantee of the order or what the arguments are going to be. In fact, you shouldn't have a constructor at all.

packages/neo-one-smart-contract-lib/src/ownership/Ownable.ts
```diff
@@ -0,0 +1,52 @@
...
+    }
+
+    @constant
+    public onlyOwner() {
+      if (this.mutableOwner && !Address.isCaller(this.mutableOwner)) {
+        throw new Error('not owner');
+      }
+    }
+
+    public protected(to: Address): boolean {
```
dicarlo2: Remove?

packages/neo-one-smart-contract-lib/src/ownership/Ownable.ts
```diff
@@ -0,0 +1,52 @@
...
+      this.mutableOwner = args[0];
+    }
+
+    @constant
+    public get owner(): Address | undefined {
+      return this.mutableOwner !== undefined ? this.mutableOwner : undefined;
+    }
+
+    @constant
+    public onlyOwner() {
```
dicarlo2: `protected`, it doesn't need to be accessible outside of the class.

packages/neo-one-smart-contract-lib/src/ownership/Ownable.ts
```diff
@@ -0,0 +1,52 @@
...
+export function Ownable<TBase extends Constructor<SmartContract>>(base: TBase) {
+  class OwnableClass extends base {
+    protected mutableOwner: Address | undefined;
+    public constructor(...args: any[]) {
+      super();
+      this.mutableOwner = args[0];
+    }
+
+    @constant
+    public get owner(): Address | undefined {
```
dicarlo2: This seems a bit awkward making it possibly `undefined`. Does it make sense to enforce there is always an owner for something that is `Ownable`?

packages/neo-one-smart-contract-lib/src/ownership/Ownable.ts
```diff
@@ -0,0 +1,52 @@
...
+      return this.mutableOwner !== undefined ? this.mutableOwner : undefined;
```
dicarlo2: `return this.mutableOwner`

packages/neo-one-smart-contract-lib/src/ownership/Ownable.ts
```diff
@@ -0,0 +1,52 @@
...
+    public transferOwnership(to: Address): boolean {
+      return this._transferOwnership(to);
+    }
+    public renounceOwnership() {
+      this.onlyOwner();
+
+      this.mutableOwner = undefined;
+    }
+    protected _transferOwnership(to: Address): boolean {
+      this.onlyOwner();
```
dicarlo2: The implementation here can just be moved to `transferOwnership`

packages/neo-one-smart-contract-lib/src/index.ts
```diff
@@ -1,4 +1,3 @@
 // tslint:disable-next-line export-name
 export { NEP5Token } from './NEP5Token';
-export { DesignatedCaller } from './ownership/DesignatedCaller';
```
dicarlo2: This commit isn't based off master - please make sure your PRs/commits are based off master.

packages/neo-one-smart-contract-lib/src/index.ts
```diff
@@ -1,4 +1,3 @@
...
-export { TransferableOwnership } from './ownership/TransferableOwnership';
+export { Ownable } from './ownership/Ownable';
```
dicarlo2: Should export from the index i.e. `export { Ownable } from './ownership'`

packages/neo-one-smart-contract-lib/src/ownership/Ownable.ts
davemneo: @dicarlo2  Thanks for spelling it out this time.  Last time I think I was bit by a thing that @afragapane and I observed: this scroll-bar hiding what you were commenting on
image
https://github.com/neo-one-suite/neo-one/pull/1057

packages/neo-one-smart-contract-lib/src/ownership/Ownable.ts
davemneo: @dicarlo2  The template for this framework has a `renounce()` that sets `_owner` to `Address(0)`.
Is it possible for address zero to make requests to the blockchain?   Otherwise, what `  Address Literal String ` would be best to use for this `renounced` state?

packages/neo-one-smart-contract-lib/src/ownership/Ownable.ts
davemneo: 👍

packages/neo-one-smart-contract-lib/src/ownership/Ownable.ts
davemneo: 👍

packages/neo-one-smart-contract-lib/src/ownership/Ownable.ts
davemneo: yeeah and sleep.
👍

packages/neo-one-smart-contract-lib/src/ownership/Ownable.ts
davemneo: Abstract it becomes.  👍

packages/neo-one-smart-contract-lib/src/ownership/Ownable.ts
davemneo: 👍

packages/neo-one-smart-contract-lib/src/ownership/Ownable.ts
davemneo: ❓ Again following the logic division as it is demonstrated in the templates we were following. 
With that said: Leave or Move?  
I did move `onlyOwner()` check tho.

packages/neo-one-smart-contract-lib/src/ownership/Ownable.ts
dicarlo2: Hmm, I think in that case if we want to support renouncing then we should leave it as is. We could set it to an address with all 0s, but I think setting it to undefined makes more sense idiomatically for typescript.

Instead, let's add a method that's like `ownerOrThrow` which returns an `Address` to cover the likely common pattern of a user expecting the contract to always have an owner.

packages/neo-one-smart-contract-lib/src/ownership/Ownable.ts
dicarlo2: Move - we don't need to translate it identically, in fact we should not translate identically in the cases where it's not idiomatic typescript.

packages/neo-one-smart-contract-lib/src/ownership/Ownable.ts
```diff
@@ -0,0 +1,37 @@
+import { Address, constant, createEventNotifier, SmartContract } from '@neo-one/smart-contract';
+
+const transferredOwnership = createEventNotifier<Address | undefined, Address>('transferred_ownership', 'from', 'to');
+
+export function Ownable<TBase extends Constructor<SmartContract>>(base: TBase) {
+  abstract class OwnableClass extends base {
+    protected mutableOwner: Address | undefined;
```
dicarlo2: Make this `private`

packages/neo-one-smart-contract-lib/src/ownership/Ownable.ts
```diff
@@ -0,0 +1,37 @@
...
+
+    @constant
+    public get owner(): Address | undefined {
+      return this.mutableOwner;
+    }
```
dicarlo2: nit: add a newline between each method.

tsconfig.json
```diff
@@ -37,11 +37,13 @@
   "exclude": [
     "dist",
     "packages/neo-one-server-plugin-wallet/src/contracts/*.ts",
-    "packages/neo-one-smart-contract-lib/src/*.ts",
     "packages/neo-one-smart-contract/src/*.d.ts",
+    "packages/neo-one-smart-contract-lib/src/index.ts",
```
dicarlo2: Why do we need the `"packages/neo-one-smart-contract-lib/src/*.ts",` change?

packages/neo-one-smart-contract-lib/src/__tests__/ownership/TestOwnableContract.test.ts
```diff
@@ -0,0 +1,80 @@
...
+        },
+      ],
+      async ({ client, networkName, testOwnable: smartContract, masterAccountID }) => {
+        crypto.addPublicKey(
+          common.stringToPrivateKey(RECIPIENT.PRIVATE_KEY),
+          common.stringToECPoint(RECIPIENT.PUBLIC_KEY),
+        );
+
+        const deployResult = await smartContract.deploy(masterAccountID.address, { from: masterAccountID });
+
```
dicarlo2: nit: remove newline - the line above is organizationally/logically related to the following lines.

packages/neo-one-smart-contract-lib/src/__tests__/ownership/TestOwnableContract.test.ts
```diff
@@ -0,0 +1,80 @@
...
+          throw new Error(deployReceipt.result.message);
+        }
+
+        expect(deployReceipt.result.gasConsumed.toString()).toMatchSnapshot('deploy consumed');
+        expect(deployReceipt.result.gasCost.toString()).toMatchSnapshot('deploy cost');
+        expect(deployReceipt.result.value).toBeTruthy();
+
+        const [initialOwner, recipient] = await Promise.all([
+          smartContract.owner(),
+
```
dicarlo2: nit: remove newline

packages/neo-one-smart-contract-lib/src/__tests__/ownership/TestOwnableContract.test.ts
```diff
@@ -0,0 +1,80 @@
...
+          client.providers.memory.keystore.addUserAccount({
+            network: networkName,
+            name: 'recipient',
+            privateKey: RECIPIENT.PRIVATE_KEY,
+          }),
+        ]);
+
+        expect(initialOwner.toString()).toEqual(masterAccountID.address);
+        expect(initialOwner.toString()).not.toEqual(recipient.userAccount.id.address);
```
dicarlo2: Seems redundant

packages/neo-one-smart-contract-lib/src/__tests__/ownership/TestOwnableContract.test.ts
```diff
@@ -0,0 +1,80 @@
...
+
+        const transferReciept = await smartContract.transferOwnership.confirmed(recipient.userAccount.id.address, {
+          from: masterAccountID,
+        });
+        const newCaller = await smartContract.owner();
```
dicarlo2: nit: `newOwner`

packages/neo-one-smart-contract-lib/src/__tests__/ownership/TestOwnableContract.test.ts
```diff
@@ -0,0 +1,80 @@
...
+        if (transferReciept.result.state !== 'HALT') {
+          throw new Error(transferReciept.result.message);
+        }
+
+        expect(newCaller.toString()).toEqual(recipient.userAccount.id.address);
+
+        const bogusTransferReciept = await smartContract.transferOwnership.confirmed(masterAccountID.address, {
```
dicarlo2: nit: `bogusTransferReceipt`.

packages/neo-one-smart-contract-lib/src/__tests__/ownership/TestOwnableContract.test.ts
```diff
@@ -0,0 +1,80 @@
...
+          from: masterAccountID,
+        });
+        expect(bogusTransferReciept.result.state).toEqual('FAULT');
+
+        const finalOwner = await smartContract.owner();
+        expect(finalOwner.toString()).toEqual(recipient.userAccount.id.address);
+
+        expect(bogusTransferReciept.result.value).toBeFalsy();
```
dicarlo2: nit: move this to right after the fault state check.

packages/neo-one-smart-contract-lib/src/__tests__/ownership/TestOwnableContract.test.ts
```diff
@@ -0,0 +1,80 @@
+import { common, crypto } from '@neo-one/client-common';
```
dicarlo2: Sidebar: these tests are great, very thorough.

packages/neo-one-smart-contract-lib/src/ownership/Ownable.ts
davemneo: 👍

tsconfig.json
davemneo: @dicarlo2  The Test......test.ts files don't contain smart contract logic, tsc errors aren't thrown so i figure this is as good an opportunity as any to un-exclude these file.

packages/neo-one-smart-contract-lib/src/ownership/Ownable.ts
davemneo: Doing so prevents:
```  public constructor(protected mutableOwner: Address = Deploy.senderAddress) {```
Unless it is optimal to take in owner and have the extending class responsible for `this.initialOwner(owner)` that has access to set the private `mutableOwner` member.

File: packages/neo-one-smart-contract-lib/src/ownership/Ownable.ts (hunk 6 above)
Comment (davemneo) [reply packages/neo-one-smart-contract-lib/src/ownership/Ownable.ts]: 👍

File: packages/neo-one-smart-contract-lib/src/ownership/Ownable.ts (hunk 12 above)
Comment (davemneo) [reply packages/neo-one-smart-contract-lib/src/ownership/Ownable.ts]: 👍

File: packages/neo-one-smart-contract-lib/src/__tests__/ownership/TestOwnableContract.test.ts (hunk 16 above)
Comment (davemneo) [reply packages/neo-one-smart-contract-lib/src/__tests__/ownership/TestOwnableContract.test.ts]: I was verifying that my addresses were not the same. 
Removed.

File: packages/neo-one-smart-contract-lib/src/__tests__/ownership/TestOwnableContract.test.ts (hunk 17 above)
Comment (davemneo) [reply packages/neo-one-smart-contract-lib/src/__tests__/ownership/TestOwnableContract.test.ts]: 👍

File: packages/neo-one-smart-contract-lib/src/__tests__/ownership/TestOwnableContract.test.ts (hunk 20 above)
Comment (davemneo) [reply packages/neo-one-smart-contract-lib/src/__tests__/ownership/TestOwnableContract.test.ts]: 👍 thnx

File: packages/neo-one-smart-contract-lib/src/__tests__/ownership/TestOwnableContract.test.ts (hunk 19 above)
Comment (davemneo) [reply packages/neo-one-smart-contract-lib/src/__tests__/ownership/TestOwnableContract.test.ts]: 👍

File: packages/neo-one-smart-contract-lib/src/__tests__/ownership/TestOwnableContract.test.ts [hunk 21]
```diff
@@ -0,0 +1,84 @@
...
+        const noOwner = await smartContract.owner();
+        expect(noOwner).toEqual(undefined);
+
+        try {
+          const ownerOrThrow = await smartContract.ownerOrThrow.confirmed({
+            from: recipient.userAccount.id,
+          });
+          expect(ownerOrThrow.result.state).toBe('FAULT');
+        } catch {
+          expect(true).toBe(true);
```
dicarlo2: Do we expect an error to be thrown here? If so, use 

```
await expect(smartContract.ownerOrThrow.confirmed({
  from: recipient.userAccount.id,
})).rejects.toMatchSnapshot();
```

packages/neo-one-smart-contract-lib/src/ownership/Ownable.ts
dicarlo2: Yea I think that's the right way to go about it - we don't want subclasses arbitrarily modifying this value.

davemneo: 👍

packages/neo-one-smart-contract-lib/src/ownership/Ownable.ts
davemneo: 👍

packages/neo-one-smart-contract-lib/src/__tests__/ownership/TestOwnableContract.test.ts
davemneo: .rejects.toBeDefined()

packages/neo-one-smart-contract-lib/src/ownership/Ownable.ts
```diff
@@ -0,0 +1,62 @@
...
+const ownershipTransferred = createEventNotifier<Address | undefined, Address>('transferred_ownership', 'from', 'to');
+
+export function Ownable<TBase extends Constructor<SmartContract>>(base: TBase) {
+  abstract class OwnableClass extends base {
+    protected abstract initialOwner: Address;
+    private mutableOwner: Address | undefined = undefined;
+    private mutableOwnerInitialized = false;
+
+    // tslint:disable-next-line:no-any
+    public constructor(...args: any[]) {
```
dicarlo2: Is there a reason the constructor was added back in?

packages/neo-one-smart-contract-lib/src/ownership/Ownable.ts
```diff
@@ -0,0 +1,62 @@
+import { Address, SmartContract, createEventNotifier } from '@neo-one/smart-contract';
+
+const ownershipTransferred = createEventNotifier<Address | undefined, Address>('transferred_ownership', 'from', 'to');
```
dicarlo2: `notifyTransferOwnership` and `transfer_ownership`.

packages/neo-one-smart-contract-lib/src/ownership/Ownable.ts
```diff
@@ -0,0 +1,62 @@
...
+    private mutableOwner: Address | undefined = undefined;
+    private mutableOwnerInitialized = false;
+
+    // tslint:disable-next-line:no-any
+    public constructor(...args: any[]) {
+      super(args);
+    }
+
+    public get owner(): Address | undefined {
+      return this.mutableOwner ? this.mutableOwner : this.initializeOwner();
```
dicarlo2: Change all of these truthy checks to be explicit `=== undefined` or `!== undefined`

packages/neo-one-smart-contract-lib/src/ownership/Ownable.ts
davemneo: Yes, testing apparatus for something I was trying, but it was left behind.  thnx.

packages/neo-one-smart-contract-lib/src/ownership/Ownable.ts
```diff
@@ -0,0 +1,57 @@
+import { Address, SmartContract, createEventNotifier } from '@neo-one/smart-contract';
+
+const notifyTransferOwnership = createEventNotifier<Address | undefined, Address>('transfer_ownership', 'from', 'to');
```
dicarlo2: the `from` `Address` can never be `undefined`, no?

packages/neo-one-smart-contract-lib/src/ownership/Ownable.ts
```diff
@@ -0,0 +1,57 @@
...
+
+      if (owner === undefined) {
+        throw new Error('no owner');
+      }
+
+      return owner;
+    }
+
+    protected onlyOwner() {
+      if (this.mutableOwner !== undefined && !Address.isCaller(this.mutableOwner)) {
```
dicarlo2: This should be using `this.ownerOrThrow` no? Seems like a bug - if there's no owner then this should throw I think.

davemneo: "bug" might be a little harsh, I'd say mine is between 4 and 12 operations more efficient and achieves same goal.

davemneo:
@danwbyrne 
> hat is the order that these PRs need to go in? I'm starting to get lost by them :P

This one first.

packages/neo-one-smart-contract-lib/src/ownership/Ownable.ts
danwbyrne: I think he is suggesting something like:
```
protected onlyOwner() {
  if (!Address.isCaller(this.ownerOrThrow())) {
    throw new Error('not owner');
  }
}
```
(edit because assignment)

danwbyrne: (because if `this.mutableOwner` is undefined then this will pass always)

packages/neo-one-smart-contract-lib/src/ownership/Ownable.ts
davemneo: After modifying `transferOwnership` to use `ownerOrThrow` otherwise, the `mutableOwner` as passed is possibly `undefined`.

packages/neo-one-smart-contract-lib/src/ownership/Ownable.ts
davemneo: @danwbyrne  yes, that was my glitch the !==undefined.  But after standup, we are favoring code readability over scrutinizing nominal amounts of gas for extra stack operations.   Change will be made, thank you!

packages/neo-one-smart-contract-lib/src/ownership/Ownable.ts
davemneo: Answer discussed @ standup, going with "Easy to read".

davemneo:
> just fix the deepscan

@danwbyrne  Unrelated

PR #1398, opened 2019-04-30
feat(rpc): Re-enable rateLimit middleware on the rpc server

danwbyrne:
Re-enable the RPC RateLimiter middleware with appropriate defaults that allow it to pass tests but also still be a failsafe if someone messed up and had a client send a TON of requests at once.


danwbyrne:
the `tooBusy` middleware still caused a trip on CI. I'll investigate that, it doesn't trip locally.

PR #1400, opened 2019-05-01
feat(contract-test-name): provide smartContractName hints

### LINKED ISSUE #1399: TokenUtils should provide a hint when smartContractName is incorrect
When testing smartContracts:  If the smartContractName provided does not match any of the keys within the submitted test options, provide a hint.

davemneo:
withContracts:
1.) Added a companion utility function to `withContracts` to test if the supplied `smartContractName` is available in the list of contracts.
2.) Rename `testToken` to `tokenTester` this seems like a reasonable rename to avoid confusing with the actual `TestToken`

If it is just an issue with matching the case of letters:
image

If the provided smart contract name is just wrong:
image

fixes #1399

### DIFF
--- a/packages/neo-one-smart-contract-test/src/withContracts.ts
+++ b/packages/neo-one-smart-contract-test/src/withContracts.ts
@@ -1,4 +1,4 @@
-import { NEOONEDataProvider } from '@neo-one/client-core';
+import { NEOONEDataProvider, SmartContractAny } from '@neo-one/client-core';
 import { createCompilerHost } from '@neo-one/smart-contract-compiler-node';
 import {
   Contract,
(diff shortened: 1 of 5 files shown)


packages/neo-one-smart-contract-test/src/withContracts.ts
```diff
@@ -10,6 +10,39 @@ import { createNode } from './createNode';
...
+// tslint:disable-next-line no-any
+export const getContractFromOptions = (options: any, smartContractName: string): SmartContractAny => {
+  const smartContract: SmartContractAny = options[smartContractName];
+
+  // tslint:disable-next-line: strict-type-predicates
+  if (smartContract === undefined) {
+    const re = new RegExp(smartContractName.toLowerCase(), 'g');
+    const matches = Object.keys(options).filter((key) => key.toLowerCase().match(re));
+    if (matches.length) {
+      // tslint:disable-next-line: no-console
```
danwbyrne: Is there a `monitor` around we could pass in and `monitor.logError(...)` instead of calling console.error directly? :thinking:

davemneo: I couldn't find any `monitor` available to the calling context:
`packages/neo-one-smart-contract-lib/src/__data__/tokenUtils.ts`

Issue #1414, opened 2019-05-24
eliminate @neo-one/types


danwbyrne:
### Description
"The global types causes issues because it leaks into the project that includes neo-one" -Alex

I'll try alternative ways of re-exporting `BN` as a util and see what I can do with the other types. 

### Steps to Reproduce

? I don't have an example

**Expected behavior:** no leaking

**Actual behavior:** leaking

PR #1416, opened 2019-05-24
fix(@neo-one/types): eliminate global types from @neo-one/types

danwbyrne:
### Description of the Change

eliminate the root declaration files in `@neo-one/types`, transition the old global types to helper types in `@neo-one/utils` and correct the files that depended on them.

### Test Plan
```
yarn test
yarn e2e -w 1
```

### Benefits

Stops it from overflowing into other projects that depend on `@neo-one`.

### Possible Drawbacks

Some more things may be broken.

### Applicable Issues

#1414


packages/neo-one-client-common/src/types.ts
```diff
@@ -1,5 +1,6 @@
 // tslint:disable deprecation no-any
 import { Monitor } from '@neo-one/monitor';
+import { Omit } from '@neo-one/utils';
```
dicarlo2: Ah I should've mentioned this - can you rename this to `OmitStrict`? TypeScript 3.5 will have an `Omit` type that's not quite the same.

yarn.lock
```diff
@@ -1891,6 +1891,45 @@
     call-me-maybe "^1.0.1"
     glob-to-regexp "^0.3.0"
 
+"@neo-one/cli@1.1.1":
```
dicarlo2: One of the package.json dependencies is wrong, can you figure out why this is getting pulled in?

PR #1417, opened 2019-05-25
feat(DAPI): Add DapiUserAccountProvider in compliance with DAPI proposal

afragapane:
### Description of the Change
Neo-one UserAccountprovider implementation for the DAPI proposal (https://github.com/neo-project/proposals/pull/69).


danwbyrne:
Is this ready for reviewing? @afragapane

afragapane:
@danwbyrne , yes, it should be in a good state for a preliminary review.

packages/neo-one-client-core/src/user/Dapi.ts
```diff
@@ -0,0 +1,396 @@
...
+  readonly confirmations: number; // Confirmation number (number of blocks after this block)
+  readonly nextblockhash: string; // Next block hash
+}
+
+export interface TransactionAttribute {
+  readonly usage: string;
+  readonly data: string;
+}
+
+export interface TransactionScript {
```
danwbyrne: same as `ScriptDetails` above

packages/neo-one-client-common/src/common.ts
```diff
@@ -26,7 +26,8 @@ export type UInt160Hex = string;
 
 const UINT160_BUFFER_BYTES = 20;
 
-const isUInt160 = (value: {}): value is UInt160 => value instanceof Buffer && value.length === UINT160_BUFFER_BYTES;
+export const isUInt160 = (value: {}): value is UInt160 =>
```
dicarlo2: These are already exported below, please revert.

packages/neo-one-client-core/src/__tests__/user/DapiUserAccountProvider.test.ts
```diff
@@ -0,0 +1,414 @@
+// import { ECPoint, Param, TransactionResult, UInt160, UInt256 } from '@neo-one/client-common';
```
dicarlo2: Should this be commented out?

packages/neo-one-client-core/src/user/Dapi.ts
```diff
@@ -0,0 +1,396 @@
...
+   * any input arguments for the operation
+   */
+  readonly args: ReadonlyArray<Argument>;
+  /**
+   * Network to submit this request to. If omitted, will default to network the wallet is currently set to.
+   */
+  readonly network?: string;
+}
+
+export interface BlockTransactionDetails {
```
dicarlo2: If we don't use this anywhere I would just leave this commented out so we don't use something typed with `any`. Either that, or add the full types in.

packages/neo-one-client-core/src/user/Dapi.ts
```diff
@@ -0,0 +1,396 @@
...
+  readonly verification: string;
+}
+
+export interface TransactionDetails {
+  readonly txid: string;
+  readonly size: number;
+  readonly type: string;
+  readonly version: number;
+  readonly attributes: ReadonlyArray<TransactionAttribute>;
+  readonly vin: ReadonlyArray<any>;
```
dicarlo2: Same here, comment these out.

packages/neo-one-client-core/src/user/Dapi.ts
```diff
@@ -0,0 +1,396 @@
...
+}
+
+export interface AssetInput {
+  readonly txid: string;
+  readonly index: number;
+}
+
+export interface AssetOutput {
+  readonly asset: string;
+  // number???
```
dicarlo2: remove?

packages/neo-one-client-core/src/user/Dapi.ts
```diff
@@ -0,0 +1,396 @@
+import { AddressString, Hash256String, NetworkType, PublicKeyString } from '@neo-one/client-common';
+
+// tslint:disable no-any
```
dicarlo2: I think this can be removed given the requested changes below.

packages/neo-one-client-core/src/user/DapiUserAccountProvider.ts
```diff
@@ -0,0 +1,618 @@
...
+import { BehaviorSubject, Observable } from 'rxjs';
+import { distinctUntilChanged } from 'rxjs/operators';
+import { clientUtils } from '../clientUtils';
+import { UnknownProviderError } from '../errors';
+import { Hash256 } from '../Hash256';
+import { Argument, AssetInput, AssetOutput, Dapi, DapiNetworks } from './Dapi';
+import { FullTransfer, Provider, UserAccountProviderBase } from './UserAccountProviderBase';
+
+const paramToArgDataType = (value: ScriptBuilderParam | undefined): Argument => {
+  if (value === undefined) {
```
dicarlo2: `undefined` is supported, this should basically just work like https://github.com/neo-one-suite/neo-one/blob/51c27baa35675740e3ecc9f426395bc127495e65/packages/neo-one-client-common/src/ScriptBuilder.ts#L115

I would also reorder the conditions here to match and put a comment in both places pointing to the other that it needs to be updated if one changes.

packages/neo-one-client-core/src/user/DapiUserAccountProvider.ts
```diff
@@ -0,0 +1,618 @@
...
+        .then(({ publicKey }) => {
+          this.updateUserAccount({
+            network: this.defaultNetworkInternal$.getValue() as NetworkType,
+            publicKey,
+            address,
+            label,
+          });
+        })
+        .catch((error) => {
+          throw error;
```
dicarlo2: What's the intention here?

packages/neo-one-client-core/src/user/DapiUserAccountProvider.ts
```diff
@@ -0,0 +1,618 @@
...
+      })
+      .catch((error) => {
+        throw error;
+      });
+
+    return currentAccount;
+  }
+
+  public async selectUserAccount(_userAccountID?: UserAccountID): Promise<void> {
+    throw new Error('Cannot select accounts for this provider.');
```
dicarlo2: Should throw coded errors

packages/neo-one-client-core/src/user/DapiUserAccountProvider.ts
```diff
@@ -0,0 +1,618 @@
...
+  public constructor({
+    dapi,
+    providers,
+  }: {
+    readonly dapi: Dapi;
+    readonly providers: { readonly [network: string]: TProvider };
+  }) {
+    super();
+    this.dapi = dapi;
+    this.providers = providers;
```
dicarlo2: I don't think you need multiple providers here, why can't we just use the same pattern as `LocalUserAccountProvider` which takes a single `Provider`?

packages/neo-one-client-core/src/user/DapiUserAccountProvider.ts
```diff
@@ -0,0 +1,577 @@
...
+        outputs: this.convertDapiOutputs(reorderOutputs(rawOutputs.concat(outputs))),
+      },
+    });
+
+    const transaction = (await provider.getTransaction(network, txid)) as InvocationTransaction;
+
+    return {
+      transaction,
+      confirmed: async () => {
+        const receipt = await provider.getTransactionReceipt(network, txid);
```
dicarlo2: `Promise.all`

packages/neo-one-client-core/src/user/DapiUserAccountProvider.ts
```diff
@@ -0,0 +1,577 @@
...
+    method: string,
+    params: ReadonlyArray<ScriptBuilderParam | undefined>,
+    _paramsZipped: ReadonlyArray<readonly [string, Param | undefined]>,
+    verify: boolean,
+    options: InvokeSendUnsafeReceiveTransactionOptions = {},
+    _sourceMaps?: Promise<SourceMaps>,
+  ): Promise<TransactionResult<RawInvokeReceipt, InvocationTransaction>> {
+    await this.initPromise;
+    const { from, networkFee, systemFee } = this.getTransactionOptions(options);
+    const network = this.defaultNetworkInternal$.getValue() as NetworkType;
```
dicarlo2: Why is this using `defaultNetworkInternal` instead of `from.network`?

I think this could be refactored a bit - all of these invoke methods should be identical (and in fact, almost are identical) between this class and `LocalUserAccountProvider` - the only difference is `invokeRaw` + the `await initPromise` at the top. I think you should be able to move all of the `invoke` methods into the base class with an abstract `invokeRaw` method and just call `await initPromise` in the implementation for this class.

packages/neo-one-client-core/src/user/DapiUserAccountProvider.ts
```diff
@@ -0,0 +1,577 @@
...
+    }
+
+    if (this.provider.iterActionsRaw === undefined) {
+      throw new NotImplementedError('iterActionsRaw');
+    }
+
+    return this.provider.iterActionsRaw(network, options);
+  }
+
+  public async transfer(
```
dicarlo2: Same comment here, try to move this into the base class and call protected abstract methods. 

Also, we can probably support multiple transfers by just calling the dapi multiple times in sequence.

packages/neo-one-client-core/src/user/DapiUserAccountProvider.ts
```diff
@@ -0,0 +1,577 @@
...
+          ...acc,
+          [balance.assetID]: new BigNumber(balance.amount),
+        }),
+        {},
+      ),
+    };
+  }
+
+  public iterBlocks(network: NetworkType, options?: IterOptions): AsyncIterable<Block> {
+    if (!this.mutableInitialized) {
```
dicarlo2: Why do we need to be initialized?

packages/neo-one-client-core/src/user/DapiUserAccountProvider.ts
```diff
@@ -0,0 +1,577 @@
...
+      throw new NotIntializedError();
+    }
+
+    return this.provider.iterBlocks(network, options);
+  }
+
+  public iterActionsRaw(network: NetworkType, options?: IterOptions): AsyncIterable<RawAction> {
+    if (!this.mutableInitialized) {
```
dicarlo2: Why do we need to be initialized?

packages/neo-one-client-core/src/user/DapiUserAccountProvider.ts
```diff
@@ -0,0 +1,577 @@
...
+    throw new NotImplementedError('updateUserAccountName');
+  }
+
+  public async getBlockCount(network: NetworkType): Promise<number> {
+    await this.initPromise;
+
+    return this.provider.getBlockCount(network);
+  }
+
+  public async getAccount(network: NetworkType, address: AddressString): Promise<Account> {
```
dicarlo2: I don't think we need to use the `dapi` here at all - just use `this.provider`

packages/neo-one-client-core/src/user/DapiUserAccountProvider.ts
```diff
@@ -0,0 +1,577 @@
...
+
+    return this.networksInternal$.getValue();
+  }
+
+  public async updateUserAccountName(_options: UpdateAccountNameOptions): Promise<void> {
+    throw new NotImplementedError('updateUserAccountName');
+  }
+
+  public async getBlockCount(network: NetworkType): Promise<number> {
+    await this.initPromise;
```
dicarlo2: Remove? This method isn't dependent on the dapi.

packages/neo-one-client-core/src/user/DapiUserAccountProvider.ts
```diff
@@ -0,0 +1,577 @@
...
+      throw new Error(`The account connected to the dapp via the dapi provider has been disconnected (logged out).`);
+    });
+
+    this.userAccounts$ = this.userAccountsInternal$;
+    this.currentUserAccount$ = this.currentAccountInternal$.pipe(distinctUntilChanged());
+    this.networks$ = this.networksInternal$;
+  }
+
+  public getCurrentUserAccount(): UserAccount | undefined {
+    if (!this.mutableInitialized) {
```
dicarlo2: I don't think we need to check if it's initialized at all, seems like it's going to just throw an error that's going to be caught and ignore anyways.

packages/neo-one-client-core/src/user/DapiUserAccountProvider.ts
```diff
@@ -0,0 +1,427 @@
...
+
+  public async updateUserAccountName(_options: UpdateAccountNameOptions): Promise<void> {
+    throw new NotImplementedError('updateUserAccountName');
+  }
+
+  public async getBlockCount(network: NetworkType): Promise<number> {
+    return this.provider.getBlockCount(network);
+  }
+
+  public async getAccount(network: NetworkType, address: AddressString): Promise<Account> {
```
dicarlo2: This can be moved to the base class

packages/neo-one-client-core/src/user/DapiUserAccountProvider.ts
```diff
@@ -0,0 +1,427 @@
...
+
+  public getNetworks(): ReadonlyArray<NetworkType> {
+    return this.networksInternal$.getValue();
+  }
+
+  public async updateUserAccountName(_options: UpdateAccountNameOptions): Promise<void> {
+    throw new NotImplementedError('updateUserAccountName');
+  }
+
+  public async getBlockCount(network: NetworkType): Promise<number> {
```
dicarlo2: This can be moved to the base class, just add `provider` as a parameter to the constructor.

packages/neo-one-client-core/src/user/DapiUserAccountProvider.ts
```diff
@@ -0,0 +1,427 @@
...
+
+  public getUserAccounts(): ReadonlyArray<UserAccount> {
+    return this.userAccountsInternal$.getValue();
+  }
+
+  public getNetworks(): ReadonlyArray<NetworkType> {
+    return this.networksInternal$.getValue();
+  }
+
+  public async updateUserAccountName(_options: UpdateAccountNameOptions): Promise<void> {
```
dicarlo2: This can be moved to the base class

packages/neo-one-client-core/src/user/DapiUserAccountProvider.ts
```diff
@@ -0,0 +1,427 @@
...
+
+  public getCurrentUserAccount(): UserAccount | undefined {
+    return this.currentAccountInternal$.getValue();
+  }
+
+  public async selectUserAccount(_userAccountID?: UserAccountID): Promise<void> {
+    throw new NotImplementedError('selectUserAccount');
+  }
+
+  public async deleteUserAccount(_userAccountID?: UserAccountID): Promise<void> {
```
dicarlo2: This can be moved to the base class

packages/neo-one-client-core/src/user/DapiUserAccountProvider.ts
```diff
@@ -0,0 +1,427 @@
...
+    this.userAccounts$ = this.userAccountsInternal$;
+    this.currentUserAccount$ = this.currentAccountInternal$.pipe(distinctUntilChanged());
+    this.networks$ = this.networksInternal$;
+  }
+
+  public getCurrentUserAccount(): UserAccount | undefined {
+    return this.currentAccountInternal$.getValue();
+  }
+
+  public async selectUserAccount(_userAccountID?: UserAccountID): Promise<void> {
```
dicarlo2: This can be moved to the base class

packages/neo-one-client-core/src/user/DapiUserAccountProvider.ts
```diff
@@ -0,0 +1,427 @@
...
+
+  public async getBlockCount(network: NetworkType): Promise<number> {
+    return this.provider.getBlockCount(network);
+  }
+
+  public async getAccount(network: NetworkType, address: AddressString): Promise<Account> {
+    return this.provider.getAccount(network, address);
+  }
+
+  public iterBlocks(network: NetworkType, options?: IterOptions): AsyncIterable<Block> {
```
dicarlo2: This can be moved to the base class

packages/neo-one-client-core/src/user/DapiUserAccountProvider.ts
```diff
@@ -0,0 +1,427 @@
...
+    return this.provider.iterBlocks(network, options);
+  }
+
+  public iterActionsRaw(network: NetworkType, options?: IterOptions): AsyncIterable<RawAction> {
```
dicarlo2: This can be moved to the base class

packages/neo-one-client-core/src/user/DapiUserAccountProvider.ts
```diff
@@ -0,0 +1,427 @@
...
+      transaction,
+      confirmed: async () => this.provider.getTransactionReceipt(from.network, txid),
+    };
+  }
+
+  public async claim(_options?: TransactionOptions): Promise<TransactionResult<TransactionReceipt, ClaimTransaction>> {
+    throw new NotImplementedError('claim');
+  }
+
+  public async invokeCompleteSend(
```
dicarlo2: Remove this method entirely and rename `invokeCompleteSendBase` to `invokeCompleteSend`

packages/neo-one-client-core/src/user/DapiUserAccountProvider.ts
```diff
@@ -0,0 +1,427 @@
...
+      params,
+      paramsZipped,
+      hash,
+      options,
+      sourceMaps,
+      this.provider,
+    );
+  }
+
+  public async invokeRefundAssets(
```
dicarlo2: Remove this method entirely and rename `invokeRefundAssetsBase` to `invokeRefundAssets`

packages/neo-one-client-core/src/user/DapiUserAccountProvider.ts
```diff
@@ -0,0 +1,427 @@
...
+    _method: string,
+    _params: ReadonlyArray<ScriptBuilderParam | undefined>,
+    _paramsZipped: ReadonlyArray<readonly [string, Param | undefined]>,
+    _options?: TransactionOptions,
+    _sourceMaps?: Promise<SourceMaps>,
+  ): Promise<TransactionResult<TransactionReceipt, ClaimTransaction>> {
+    throw new NotImplementedError('invokeClaim');
+  }
+
+  public async call(
```
dicarlo2: This can be moved to the base class. There's no need to `await this.initPromise`

packages/neo-one-client-core/src/user/DapiUserAccountProvider.ts
```diff
@@ -0,0 +1,427 @@
...
+    transfers = [],
+    rawInputs = [],
+    rawOutputs = [],
+    scripts,
+    onConfirm,
+    reorderOutputs = (outs) => outs,
+  }: InvokeRawOptions<T>): Promise<TransactionResult<T, InvocationTransaction>> {
+    await this.initPromise;
+    const { from, networkFee, systemFee } = this.getTransactionOptions(options);
+    if (invokeMethodOptions === undefined) {
```
dicarlo2: This is never undefined, make the type required.

packages/neo-one-client-core/src/user/DapiUserAccountProvider.ts
```diff
@@ -0,0 +1,427 @@
...
+      throw new InvalidArgumentError('InvokeMethodOptions', 'invokeMethodOptions', invokeMethodOptions);
+    }
+    const { contract, invokeMethod: method, params } = invokeMethodOptions;
+
+    const { gas } = await this.checkSystemFees({
```
dicarlo2: Attributes need to be used identically to `LocalUserAccountProvider`. Use the `txHashAttributes` property of `dapi.invoke`

packages/neo-one-client-core/src/user/DapiUserAccountProvider.ts
```diff
@@ -0,0 +1,427 @@
...
+      args: params.map(paramToArgDataType),
+      network: this.defaultNetworkInternal$.getValue(),
+      triggerContractVerification: verify,
+      assetIntentOverrides: {
+        inputs: this.convertDapiInputs(rawInputs.concat(inputs)),
+        outputs: this.convertDapiOutputs(reorderOutputs(rawOutputs.concat(outputs))),
+      },
+    });
+
+    const transaction = (await this.provider.getTransaction(from.network, txid)) as InvocationTransaction;
```
dicarlo2: Rather than cast, just check that the transaction type is an invocation transaction, and throw otherwise.

packages/neo-one-client-core/src/user/DapiUserAccountProvider.ts
```diff
@@ -0,0 +1,427 @@
...
+    });
+
+    const { inputs, outputs } = await this.getTransfersInputOutputs({
+      transfers,
+      from,
+      gas: networkFee.plus(gas),
+      provider: this.provider,
+    });
+
+    const { txid } = await this.dapi.invoke({
```
dicarlo2: It seems like the only concrete differences between this method and the equivalent in `LocalUserAccountProvider` are line 309 and 338-350 vs. the `sendTransaction` bit in `LocalUserAccountProvider`.

packages/neo-one-client-core/src/user/DapiUserAccountProvider.ts
afragapane: I did it this way for the cases in @neo-one/client-full-core/LocalUserAccountProvider.ts.  There seem to be cases where the script needs to be passed without contract/invocation details.

afragapane:
Moving this to 2 new PRs - LocalUserAccountProvider refactor + DapiUserAccountProvider

Issue #1431, opened 2019-06-05
Ora Spinner "spazzes out" on neo-one startup


danwbyrne:
### Description

Our ora-spinner is messed up somehow right now, need to take another look.

### Steps to Reproduce

```
yarn build
node ./dist/neo-one/packages/neo-one-cli/bin/neo-one version
```

after running that once you will have to run:
```
node ./dist/neo-one/packages/neo-one-cli/bin/neo-one reset --static-neo-one
```
between runs to see if the error is resolved or not.

**Expected behavior:** The spinner title should be either "Starting NEO ONE server" or "Checking NEO ONE version", but not both at the same time.
**Actual behavior:** It rapidly flashes between 2 messages.

**Reproduces how often:** every time it needs to start the server to do a check.

### Versions

latest

PR #1461, opened 2019-06-13
fix(cli): fix cli-spinner title flashing

danwbyrne:
### Description of the Change

Fix the flashing of the CLI spinner by adding a server command to check for a newer version and using that, instead of what we did before which had the versionSpinner not being killed until *after* the server spinner had started.

### Test Plan

```
yarn build
node ./dist/neo-one/packages/neo-one-cli/bin/neo-one version
```
or any other command and note how they don't overlap anymore.

I also moved some of the logic/text around to make it more clear when it is starting/checking for a running server.

### Alternate Designs

I'm sure there are several other ways to organize this.

### Benefits

Less awful to look at.

### Applicable Issues

#1431

### DIFF
--- a/packages/neo-one-cli/src/InteractiveCLI.ts
+++ b/packages/neo-one-cli/src/InteractiveCLI.ts
@@ -268,47 +268,37 @@ export class InteractiveCLI {
             }
             manager = new ServerManager({ dataPath, serverVersion: VERSION });
 
-            // tslint:disable-next-line no-any
-            let spinner: any | undefined;
-            // tslint:disable-next-line no-any
-            let verSpinner: any | undefined;
-
             if (first) {
-              verSpinner = ora(`Checking for @neo-one/cli updates...`).start();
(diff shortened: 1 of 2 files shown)


packages/neo-one-cli/src/InteractiveCLI.ts
```diff
@@ -268,47 +268,33 @@ export class InteractiveCLI {
...
 
-            // tslint:disable-next-line no-any
-            let spinner: any | undefined;
-            // tslint:disable-next-line no-any
-            let verSpinner: any | undefined;
-
             if (first) {
-              verSpinner = ora(`Checking for @neo-one/cli updates...`).start();
+              const versionSpinner = ora(`Checking for @neo-one/cli updates...`).start();
+              const newVersion = await manager.getNewerServerVersion();
```
dicarlo2: Needs to be in a `try`/`catch` to end the spinner on error.

Issue #1447, opened 2019-06-05
Update cypress tests to use production version of the website


dicarlo2:

dicarlo2:
This line[1] needs to be replaced with logic to build the production version + launch the server(s) to serve the production website

[1] https://github.com/neo-one-suite/neo-one/blob/master/scripts/website/runCypress.ts#L89

PR #1467, opened 2019-06-18
fix(cypress): update cypress tests and website

spencercorwin:
### Description of the Change

I updated the way that `yarn test:cypress` builds the production website and runs the website servers so that Cypress can be run against the production website. I adjusted the way that Cypress executes the tests so that they should pass the tests without running into problems with the website's editor. I tweaked several tests to match what the current version of the website was producing. I also added a plugin called `cypress-plugin-retries` that should re-run tests that fail 3 times before actually failing the test. This functionality is not currently native to Cypress itself.

I also added the ability to run `yarn test:cypress-express` which will add an `--express` flag to `yarn test:cypress` which tells the script to do the same thing except only run a single test. I've chosen Lesson 2 Chapter 5 for this express test because it's quick but also tests a range of editor and NEO-ONE functionality.

Fixes: #1447 

### Test Plan

`yarn test:cypress` or `yarn test:cypress-express`

Note that the normal tests take 2 hours and can run your local machine hard. Cypress will save videos and screenshots of the tests into `neo-one/cypress/videos` and `neo-one/cypress/screenshots` as it runs the tests so you can see what happened in headless mode and diagnose failures. I have detailed notes and saved videos if needed to cross-check results

### Alternate Designs

None.

### Benefits

Single command for E2E testing of NEO ONE website courses and eventual CI integration

### Possible Drawbacks

The tests take 2 hours to run and this is only testing the courses, unless run in *express* mode.

### Applicable Issues

The tests take 2 hours to run. The bulk of the time is from Cypress typing in the solutions into the editor. Cypress doesn't currently support copy/paste functionality although it's been requested and they say they're working on it. Cypress does allow direct DOM manipulation but our editor is not a simple `<textarea>`. Each line of the editor is a series of nested `div`s. So it's hard for me to see any way around having Cypress type in the text. Cypress does support "Parallelization", allowing us to run the tests in parallel on multiple virtual machines but using the same test environment, which could mitigate this problem in CI.


scripts/website/runCypress.ts
```diff
@@ -46,22 +46,22 @@ const runCypress = async ({ report, coverage }: { readonly report: boolean; read
...
-        process.exit(exitCode);
-      })
-      .catch((error) => {
-        console.error(error);
-        process.exit(1);
-      });
+    try {
+      console.log(`Shutdown called with exit code ${exitCode}`);
+      mutableCleanup.forEach((callback) => callback());
+      // process.exit(exitCode);
```
spencercorwin: @dicarlo2 Uncommenting this line, `process.exit()`, will cause the two `execa` processes to *not* cancel. In other words, when the tests finish the website and servers would still be up if `process.exit()` was there. That seems strange. Is there a reason that `process.exit()` needs to be called here? Or can it be left out?

cypress/common/lessons.ts
```diff
@@ -283,33 +286,64 @@ export const lesson3 = ({
...
+      });
+    } else {
+      runTests({
+        passing: 1,
+        failing: 0,
+        suites: [
+          {
+            basename: 'utils.test.ts',
+            dirname: 'src/__tests__',
+            passing: 1,
```
dicarlo2: looks like this is the only field that's different right? Might be better to do something like `chapter === 3 || ... ? 2 : 1`

scripts/website/runCypress.ts
dicarlo2: I think you don't want to change this bit of code - let's chat about it today.

spencercorwin: Ok. Ready when you are

package.json
```diff
@@ -123,15 +123,15 @@
...
-    "website:start-preview-prod": "cross-env NEO_ONE_CACHE=true NEO_ONE_PROD=true yarn run ts-node ./scripts/website/compile --watch --bundle preview",
+    "website:start-workers-prod": "cross-env NEO_ONE_CACHE=true NEO_ONE_PROD=true yarn run ts-node ./scripts/website/compile --bundle workers",
+    "website:start-overlay-prod": "cross-env NEO_ONE_CACHE=true NEO_ONE_PROD=true yarn run ts-node ./scripts/website/compile --bundle overlay",
+    "website:start-tools-prod": "cross-env NEO_ONE_CACHE=true NEO_ONE_PROD=true yarn run ts-node ./scripts/website/compile --bundle tools",
+    "website:start-preview-prod": "cross-env NEO_ONE_CACHE=true NEO_ONE_PROD=true yarn run ts-node ./scripts/website/compile --bundle preview && yarn run serve dist/preview -p 8080",
     "website:start-testRunner-prod": "cross-env NEO_ONE_CACHE=true NEO_ONE_PROD=true yarn run ts-node ./scripts/website/compile --bundle testRunner && yarn run serve dist/testRunner -p 8081",
     "website:start-server-prod": "cross-env NEO_ONE_CACHE=true NEO_ONE_PROD=true yarn run ts-node ./scripts/website/compile --bundle server && yarn run serve dist/server -p 3001",
```
spencercorwin: Compare this to `website:start-server`

package.json
```diff
@@ -123,15 +123,15 @@
     "website:start-static": "cross-env NODE_OPTIONS=\"--max-old-space-size=6144\" TS_NODE_PROJECT=tsconfig/tsconfig.es2017.cjs.json yarn run react-static start",
```
spencercorwin: The `website:start-prod` script uses line 122 above this line for the server. Compare to the `website:start-server-prod` script below. The `website:start-server-prod` script doesn't seem to work. Using the `serve` package just makes the compiled files available. Adding the `--watch` flag like this command does works though.

package.json
```diff
@@ -120,18 +120,19 @@
     "website:start-preview": "yarn run ts-node ./scripts/website/compile --watch --bundle preview",
     "website:start-testRunner": "yarn run ts-node ./scripts/website/compile --watch --bundle testRunner",
     "website:start-server": "yarn run ts-node ./scripts/website/compile --watch --bundle server",
+    "website:start-server-test": "cross-env NEO_ONE_CACHE=true NEO_ONE_PROD=true yarn run ts-node ./scripts/website/compile --bundle server",
```
spencercorwin: The `website:start-prod` script uses line 122 above this line for the server. Compare to the `website:start-server-prod` script below. The `website:start-server-prod` script doesn't seem to work. Using the `serve` package just makes the compiled files available. Adding the `--watch` flag like this command does works though.

package.json
```diff
@@ -120,18 +120,19 @@
...
-    "website:start-preview-prod": "cross-env NEO_ONE_CACHE=true NEO_ONE_PROD=true yarn run ts-node ./scripts/website/compile --watch --bundle preview",
+    "website:start-workers-prod": "cross-env NEO_ONE_CACHE=true NEO_ONE_PROD=true yarn run ts-node ./scripts/website/compile --bundle workers",
+    "website:start-overlay-prod": "cross-env NEO_ONE_CACHE=true NEO_ONE_PROD=true yarn run ts-node ./scripts/website/compile --bundle overlay",
+    "website:start-tools-prod": "cross-env NEO_ONE_CACHE=true NEO_ONE_PROD=true yarn run ts-node ./scripts/website/compile --bundle tools",
+    "website:start-preview-prod": "cross-env NEO_ONE_CACHE=true NEO_ONE_PROD=true yarn run ts-node ./scripts/website/compile --bundle preview && yarn run serve dist/preview -p 8080",
     "website:start-testRunner-prod": "cross-env NEO_ONE_CACHE=true NEO_ONE_PROD=true yarn run ts-node ./scripts/website/compile --bundle testRunner && yarn run serve dist/testRunner -p 8081",
     "website:start-server-prod": "cross-env NEO_ONE_CACHE=true NEO_ONE_PROD=true yarn run ts-node ./scripts/website/compile --bundle server && yarn run serve dist/server -p 3001",
```
spencercorwin: @danwbyrne The `website:start-prod` script uses the `website:start-server-test` script instead of the `website:start-server-prod` script. The `website:start-server-prod` script doesn't seem to work. Using the `serve` package just makes the compiled files available. Adding the `--watch` flag like this command does works though. I'm not sure what's the `serve` package is doing or if this is the correct way to handle this. All the other `website:start-prod` scripts use the `serve` package successfully it seems.

cypress/support/index.js
```diff
@@ -8,3 +10,7 @@ if (Cypress.env('coverageDir')) {
     });
   });
 }
+
+Cypress.on('uncaught:exception', (err, runnable) => {
```
dicarlo2: Remove

Issue #1458, opened 2019-06-10
.rc file settings are not mapped in node-protocol/Node.js


hal0x2328:
### Description

I have a .neo_one_noderc file specifying the following options:
```
  "options": {
    "consensus": {
      "enabled": true,
      "options": {
        "privateKey": "e35fa5d1652c4c65e296c86e63a3da6939bc471b741845be636e2daa320dc770",
        "privateNet": true
      }
    }
  }
```
However, I was unable to get the node to start in consensus mode at all despite it being enabled in the .rc options.

I traced the consensus initialization to `node-protocol/Node.js`, where the default values set here:
```
        const defaultOptions = {
            enabled: false,
            options: { privateKey: 'unused', privateNet: false },
        }
```
are supposed to be populated with the .rc options in the subsequent code (I think), but in my case this doesn't work, and consensus start is skipped because the consensus.enabled remains `false`.

I'm not sure if this is a bug in mapping the .rc settings to the options state or a problem with my assumptions about the .rc file configuration, however, if I hard-code the same settings from the RC file directly into the `defaultOptions` object, the node starts in consensus mode without any issue, and no other errors are reported. My other .rc file section settings such as http port, standbyValidators etc. are all honored.

### Steps to Reproduce

1. Set up a .neo_one_noderc file for a custom privatenet with 1 validator
2. Try and start the node in consensus mode

**Expected behavior:** 
The node will start in consensus mode and begin producing blocks

**Actual behavior:** 
The node starts but does not enter consensus mode, and no blocks are produced except for the genesis block

**Reproduces how often:** 
100%

### Versions
neo-one/node-protocol v1.1.3
Mac OS Mojave 10.14.3
Electron v5.0.2

### Additional Information

Also using the following .rc file settings which are imported correctly as far as I can tell:
```
  "settings": { 
    "type": "test", 
    "privatenet": true,
    "standbyValidators": ["0248be3c070df745a60f3b8b494efcc6caf90244d803a9a72fe95d9bae2120ec70"]
  }
```

hal0x2328:
A privatenet started through the cli interface works fine without any changes to Node.js, but when I copy the cli-generated configuration over to ~/.neo_one_noderc and start the node using `node-bin/bin/neo-one-node` it loads the config but doesn't even output any console logs like my previous .rc file did, it just hangs. 

I'm going to assume the .rc system is not compatible with the latest config formats and just go with the cli startup method.

dicarlo2:
[...]

PR #1478, opened 2019-06-21
chore(node): re-organize '@neo-one/node-bin' configuration layout

### LINKED ISSUE #1102: Change Node Configuration Layout
title its bad right now

danwbyrne:
organize the layout to be more ergonomic instead of what is convenient for the code

BREAKING CHANGE: old node-bin configs/.rc will almost surely not work. While I update the documentation you can refer here for a rough draft of the new layout: https://gist.github.com/danwbyrne/c6ee99309fea79d90ac3044c0e6c25fe

`haltAndBackup` re-renamed to `haltOnSync`. I understand we named it `haltAndBackup` because that is what we always do with it but it actually doesn't have to backup so this makes more sense semantically.

fix #1102 #1458


packages/neo-one-node-bin/src/utils/extractConfigurations.ts
```diff
@@ -0,0 +1,27 @@
...
+import { NetworkEnvironment, NetworkOptions } from '@neo-one/node-network';
+import { NodeEnvironment, NodeOptions } from '@neo-one/node-protocol';
+import _ from 'lodash';
+
+const envKeys = {
+  rpc: ['http', 'https', 'splashScreen'] as const,
+  backup: ['tmpPath', 'readyPath'] as const,
+  node: ['externalPort'] as const,
+  network: ['listenTCP'] as const,
+};
```
dicarlo2: Just make the whole thing `as const`

PR #1482, opened 2019-06-25
feat(cli): 'yarn neo-one init' finds framework, warns otherwise

### LINKED ISSUE #1382: Automatically detect when create-react-app or angular/vue equivalents are used and initialize appropriately
### Description

`yarn neo-one init` should be smarter

### Steps to Reproduce

1. `npx create-react-app token --typescript`
2. add neo-one deps
3. `yarn neo-one init`

**Expected behavior:** User is ready to go

**Actual behavior:** Additional configuration is required.

### Versions

v1.1.2

### Additional Information

Tutorial docs will need updating: https://neo-one.io/tutorial#Setup-for-the-Tutorial

danwbyrne:
### Description of the Change

Check the package.json of the current working directory for key dependencies in order to deduce the project framework.

Additional do some `Set` wrapping for the tsconfig.exclude overrides we do. Previously if someone did `yarn neo-one init` multiple times it would keep adding the same file to the end of the exclude. Fixed now :)

### Test Plan

kind of complicated but if you do the various `create-x-app`s you can see that the dependencies I check for are the ones generated by the frameworks. You can quickly test it works for react using
```
yarn build
node ./dist/neo-one/packages/neo-one-cli/bin/neo-one init
```
and see that the `.onerc` has "react" as its framework.

### Alternate Designs

It looks like only angular sets up additional `.json` files that could be used to identify it as angular, but it also looks like that name can be variable. I'm open to other suggestions of checking the framework.

### Benefits

quicker to hit the ground running with `@neo-one`

### Applicable Issues

fix #1382

### DIFF
--- a/packages/neo-one-server-plugin-project/src/initCommand.ts
+++ b/packages/neo-one-server-plugin-project/src/initCommand.ts
@@ -65,16 +65,20 @@ export const initCommand = ({ cli }: InteractiveCLIArgs) =>
       initDataFile({ folder: cwd, name: '.onerc', data: JSON.stringify(relativeConfig), cli }),
     ]);
 
+    if (projectConfig.codegen.framework === 'none') {
+      cli.print('Unable to determine project framework, please specify it in .onerc');
+    }
+
     const rootTsConfigPath = path.resolve(cwd, 'tsconfig.json');
     const exists = await fs.pathExists(rootTsConfigPath);
     if (exists) {
       const file = await jsonFile(rootTsConfigPath);
       const currentExcludeIn = await file.get('exclude');
-      const currentExclude = currentExcludeIn === undefined ? [] : currentExcludeIn;
+      const currentExclude = new Set<string>(currentExcludeIn);
       file.set({
-        exclude: currentExclude.concat([
-          normalizePath(path.relative(cwd, path.join(projectConfig.paths.contracts, '*.ts'))),
-        ]),
+        exclude: [
+          ...currentExclude.add(normalizePath(path.relative(cwd, path.join(projectConfig.paths.contracts, '*.ts')))),
+        ],
       });
       await file.save();
       cli.print('Added contracts directory to root tsconfig exclude.');
--- a/packages/neo-one-server-plugin-project/src/utils/loadProjectConfig.ts
+++ b/packages/neo-one-server-plugin-project/src/utils/loadProjectConfig.ts
@@ -1,17 +1,27 @@
+import { CodegenFramework } from '@neo-one/smart-contract-codegen';
 import convict from 'convict';
 import cosmiconfig from 'cosmiconfig';
-import * as path from 'path';
+import fs from 'fs-extra';
+import path from 'path';
 import { ProjectConfig, projectConfigSchema } from '../types';
 
-const validateConfig = (rootDir: string, configIn: cosmiconfig.Config): ProjectConfig => {
+const validateConfig = async (rootDir: string, configIn: cosmiconfig.Config): Promise<ProjectConfig> => {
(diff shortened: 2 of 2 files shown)


packages/neo-one-server-plugin-project/src/utils/loadProjectConfig.ts
```diff
@@ -1,17 +1,24 @@
...
 import { ProjectConfig, projectConfigSchema } from '../types';
 
-const validateConfig = (rootDir: string, configIn: cosmiconfig.Config): ProjectConfig => {
+const validateConfig = async (rootDir: string, configIn: cosmiconfig.Config): Promise<ProjectConfig> => {
   const config = convict<ProjectConfig>(projectConfigSchema);
   config.load(configIn);
   config.validate({ allowed: 'warn' });
 
   const validatedConfig = config.getProperties();
+  const framework = await getProjectFramework(rootDir);
```
dicarlo2: This will always overwrite - this should only be called (and added) if the config is not present.

packages/neo-one-server-plugin-project/src/utils/loadProjectConfig.ts
```diff
@@ -26,3 +33,35 @@ export const loadProjectConfig = async (rootDir: string): Promise<ProjectConfig>
...
+  const pkgPath = path.resolve(rootDir, 'package.json');
+  const pkgExists = await fs.pathExists(pkgPath);
+  if (!pkgExists) {
+    return 'none' as const;
+  }
+  const pkg = await fs.readJSON(path.resolve(rootDir, 'package.json'));
+  const dependencies = getPkgDependencies(pkg);
+
+  if (dependencies.has('react') || dependencies.has('react-dom')) {
+    return 'react' as const;
```
dicarlo2: Remove all of the `as const` - the explicit return type takes care of this already.

packages/neo-one-server-plugin-project/src/utils/loadProjectConfig.ts
```diff
@@ -26,3 +33,35 @@ export const loadProjectConfig = async (rootDir: string): Promise<ProjectConfig>
...
+const getProjectFramework = async (rootDir: string): Promise<CodegenFramework> => {
+  const pkgPath = path.resolve(rootDir, 'package.json');
+  const pkgExists = await fs.pathExists(pkgPath);
+  if (!pkgExists) {
+    return 'none' as const;
+  }
+  const pkg = await fs.readJSON(path.resolve(rootDir, 'package.json'));
+  const dependencies = getPkgDependencies(pkg);
+
+  if (dependencies.has('react') || dependencies.has('react-dom')) {
```
dicarlo2: Does this work with create-react-app? I don't recall if react is explicitly a dependency or just cra is.

packages/neo-one-server-plugin-project/src/initCommand.ts
```diff
@@ -65,16 +65,20 @@ export const initCommand = ({ cli }: InteractiveCLIArgs) =>
       initDataFile({ folder: cwd, name: '.onerc', data: JSON.stringify(relativeConfig), cli }),
     ]);
 
+    if (projectConfig.codegen.framework === 'none') {
+      cli.print('Unable to determine project framework, please specify it yourself in .onerc');
```
dicarlo2: drop "yourself"

packages/neo-one-server-plugin-project/src/utils/loadProjectConfig.ts
danwbyrne: Yes when I did
```
npx create-react-app token --typescript
```
react and react-dom were explicit dependencies.

PR #1507, opened 2019-07-03
refactor(LocalUserAccountProvider): Separate LocalUserAccountProvider

afragapane:
Separate LocalUserAccountProvider into abstract base class and higher order classes.

### Description of the Change
LocalUserAccountProvider refactored into a base class and higher order classes.  The base class should take care of all setup and common functionality.  The higher order class takes care of actually executing transactions & invocations.  

### Test Plan
LocalUserAccountProvider tests still pass

### Benefits
Simplifies work for DAPI and any other providers we add in the future

### DIFF
--- a/packages/neo-one-client-core/src/errors.ts
+++ b/packages/neo-one-client-core/src/errors.ts
@@ -154,3 +154,8 @@ export const InvalidMasterPathError = makeErrorWithCode(
   'INVALID_HD_MASTER_PATH',
   (path: readonly number[]) => `Invalid masterPath returned by HDStore: ${path}`,
 );
+
+export const NotImplementedError = makeErrorWithCode(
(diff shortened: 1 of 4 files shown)


dicarlo2:
LGTM. 

One thing that seems a bit strange though and I'm not sure about is it seems like we have sort of a diamond pattern going on here of

```
                <method>
      /                           \
invokeRawScript                 invokeRawMethod
      \                           /
                invokeRaw
      /                           \
executeInvokeScript               executeInvokeMethod
```

Perhaps it'll make more sense in the next PR, but it feels like it should be something more like

```
                <method>
      /                           \
invokeRawScript                 invokeRawMethod
      |                            |
executeInvokeScript             executeInvokeMethod
```

PR #1523, opened 2019-07-08
Remove change-case

### LINKED ISSUE #1449: Update neo-one playground

### LINKED ISSUE #1499: Remove `change-case` library
### Description

1. Remove `change-case` from all package.json's.
2. Look up where `pascal` and `camel` imports come from.
3. Copy those modules into `packages/neo-one-utils`
4. Cover those new utils 100%
5. ???
6. Profit

spencercorwin:
### Description of the Change

Remove the `change-case` dependency from `neo-one` by replicating the needed functionality in `@neo-one/utils`. Includes 100% test coverage. `camel` and `pascal` were used in `neo-one-smart-contract-code-gen` and `neo-one-smart-contract-test-common`. I removed `change-case` from both of those and replaced it with the `@neo-one/utils` implementations.

Fixes #1499 

### Test Plan

To test run: `yarn jest packages/neo-one-utils/src/__tests__`
Can also run: `yarn jest packages/neo-one-smart-contract-codegen/src/__tests__` and `yarn jest packages/neo-one-smart-contract-codegen/src/__tests__` to check that the affected packages pass tests.

### Alternate Designs

None. Basically copy pasted from the `change-case` source code then tweaked it to add types and remove unnecessary `null` checks.

### Benefits

No more `change-case`

### Possible Drawbacks

None.

### Applicable Issues

None.


packages/neo-one-utils/src/__tests__/changeCase.test.ts
```diff
@@ -0,0 +1,183 @@
+import { inspect } from 'util';
+import { camel, lowerCase, noCase, pascal, upperCase, upperCaseFirstNative } from '../index';
+
+const NO_CASE_TESTS: ReadonlyArray<ReadonlyArray<string>> = [
```
danwbyrne: It might have auto-completed this but we ts 3.5.2 now.

```ts
const NO_CASE_TESTS: ReadonlyArray<readonly [string, string]> = [...]
```

packages/neo-one-utils/src/__tests__/changeCase.test.ts
```diff
@@ -0,0 +1,183 @@
...
+  // https://github.com/blakeembrey/change-case/issues/21
+  ['amazon s3 data', 'amazon s3 data'],
+  ['foo_13_bar', 'foo 13 bar'],
+];
+
+describe('no case', () => {
+  NO_CASE_TESTS.forEach((test) => {
+    const before = test[0];
+    const after = test[1];
+    const replacement = test[3];
```
danwbyrne: this is just `undefined` everywhere no? Can probably remove it.

packages/neo-one-utils/src/changeCase.ts
```diff
@@ -0,0 +1,145 @@
...
+  let nextStr = String(str);
+
+  if (lang) {
+    nextStr = str.replace(lang.regexp, (m) => lang.map[m]);
+  }
+
+  return nextStr.toUpperCase();
+};
+
+export const noCase = (str: string, locale?: Locale, replacement?: string): string => {
```
danwbyrne: basically a nit: I see in some of the tests a lot of `(arg1, undefined, arg2)` which, from what I've gathered in our codebase, isn't a pattern we want to use. Could you refactor this as:
```ts
interface NoCaseOptions {
    readonly locale?: Locale;
    readonly replacement?: string;
}
export const noCase = (str: string, options: NoCaseOptions): string => {...}
```

this way we can avoid having `undefined` as an argument to noCase.

packages/neo-one-utils/src/__tests__/changeCase.test.ts
```diff
@@ -0,0 +1,183 @@
...
+  it('should upper case the first character a string', () => {
+    expect(upperCaseFirstNative('test')).toEqual('Test');
+    expect(upperCaseFirstNative('TEST')).toEqual('TEST');
+    expect(upperCaseFirstNative('tEST')).toEqual('TEST');
+    expect(upperCaseFirstNative('Test')).toEqual('Test');
+  });
+});
+
+describe('lower case', () => {
+  it('should lower case a string', () => {
```
danwbyrne: can you change all the `it`s to `test`s? We are a `test` family here @ neo-one

packages/neo-one-utils/src/changeCase.ts
```diff
@@ -0,0 +1,145 @@
...
+  // Replace periods between numeric entities with an underscore.
+  if (!mergeNumbers) {
+    result = result.replace(/ (?=\d)/g, '_');
+  }
+
+  // Replace spaces between words with an upper cased character.
+  return result.replace(/ (.)/g, (_m, $1) => upperCase($1, locale));
+};
+
+export const pascal = (value: string, locale?: Locale, mergeNumbers?: boolean): string =>
```
danwbyrne: You can do a similar refactor like I described above here.

A really good rule of thumb to have in general is functions should have *AT MOST* 2 arguments. Sometimes we get away with breaking that rule for short utility functions but its good practice to refactor any number of arguments >2 into 1 options object.

PR #1535, opened 2019-07-11
fix(@neo-one/website): Fix website editor errors

### LINKED ISSUE #1491: Monaco Editor Default Library Incorrect
### Description

The Monaco editor is using the wrong tsconfig settings by default. I have tried NUMEROUS ways of resolving this, namely:
1. I noticed `ts.ScriptTarget.ESNext` and `monaco.languages.typescript.ScriptTarget.ESNext` in `packages/neo-one-editor/languages.ts` had a mismatch in values. So I tried using the monaco value instead to no avail.
2. We do some conversion for languageIDs between 'typescript' and 'typescript-internal', tried messing around with these since I only see 'typescript' used in monaco examples but I assume it is converted before being created so problem not the issue.
3. adding `/// <reference lib="es2017" />` fixes the error, but `/// <reference lib="esnext" />` does not so naturally I just tried changing the target options listed in (1) to ES2017 which didn't fix anything.
4. following (3) I just tried setting the `lib` option to `["esnext"]` as well as `["es2017"]` but that wasn't enough either.
5. There is this command `monaco.languages.typescript.typescriptDefaults.setCompilerOptions(...)` but anywhere I tried to use it when running the website it would just say the function is undefined. I assume this has something to do with how we just use `/// <reference lib="monaco/editor" />` instead of actually importing it?

This issue first presents itself when we upgrade to typescript 3.5.2 after going back through a few different commits in the history so that is an obvious red flag. As of 0.17.1 they should be on ts 3.5 🤷‍♂ 

### Steps to Reproduce

```
yarn install
yarn website:start:dev-builds
yarn website:start:dev (other window)
```
editor has broken ts errors that don't present themselves when running/building.

**Expected behavior:** No errors like on the website now

**Actual behavior:** errors on all the editors

### Versions

current

### Additional Information

I've looked at this for several hours, don't want to keep throwing time at it.

dicarlo2:
Fix #1491 #1498

### Requirements

### DIFF
--- a/packages/neo-one-editor/src/engine/main/initializeFileSystem.ts
+++ b/packages/neo-one-editor/src/engine/main/initializeFileSystem.ts
@@ -52,6 +52,8 @@ import lib_es2017_string_d_ts from '!raw-loader!../../../../../node_modules/type
 // @ts-ignore
 import lib_es2017_typedarrays_d_ts from '!raw-loader!../../../../../node_modules/typescript/lib/lib.es2017.typedarrays.d.ts';
 // @ts-ignore
+import lib_es2018_asynciterable_d_ts from '!raw-loader!../../../../../node_modules/typescript/lib/lib.es2018.asynciterable.d.ts';
+// @ts-ignore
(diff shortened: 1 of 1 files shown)


spencercorwin:
bors retry

PR #1538, opened 2019-07-11
feat(DapiUserAccountProvider): Implement DapiUserAccountProvider

afragapane:
### Description of the Change
Implement DAPIUserAccountProvider in compliance with the the DAPI NEP (https://github.com/neo-project/proposals/pull/69)

### Test Plan
Added unit tests similar to those for LocalUserAccountProvider.

### Applicable Issues
#1497

### DIFF
--- a/packages/neo-one-client-core/src/user/DapiUserAccountProvider.ts
+++ b/packages/neo-one-client-core/src/user/DapiUserAccountProvider.ts
@@ -1,0 +1,359 @@
+import {
+  AddressString,
+  addressToScriptHash,
+  Attribute,
+  ClaimTransaction,
+  common,
+  ContractTransaction,
+  Input,
+  InvocationTransaction,
+  NetworkType,
+  Output,
+  ParamToCallbacks,
+  PublicKeyString,
+  ScriptBuilderParam,
+  TransactionReceipt,
+  TransactionResult,
+  Transfer,
+  UserAccount,
+  UserAccountID,
+  UserAccountProvider,
+} from '@neo-one/client-common';
+import BigNumber from 'bignumber.js';
+import { BehaviorSubject, Observable } from 'rxjs';
+import { distinctUntilChanged } from 'rxjs/operators';
+import { InvocationCallError, NothingToTransferError, NotImplementedError } from '../errors';
+import {
+  ACCOUNT_CHANGED,
+  Argument,
+  AssetInput,
+  AssetOutput,
+  Dapi,
+  DapiAccount,
+  DapiError,
+  DapiNetworks,
+  NETWORK_CHANGED,
+  TxHashAttribute,
+} from './Dapi';
+import {
+  ExecuteInvokeClaimOptions,
+  ExecuteInvokeMethodOptions,
+  ExecuteInvokeScriptOptions,
+  Provider,
+  UserAccountProviderBase,
+} from './UserAccountProviderBase';
+
+const paramToArgDataTypeCallbacks: ParamToCallbacks<Argument> = {
+  undefined: () => ({
+    type: 'ByteArray',
+    value: Buffer.alloc(0, 0),
+  }),
+  array: (param) => ({
+    type: 'Array',
+    value: param.map(paramToArgDataType),
+  }),
+  map: (param) => ({
+    type: 'Array',
+    value: Array.from(param).map(([k, v]) => [paramToArgDataType(k), paramToArgDataType(v)]),
+  }),
+  uInt160: (param) => ({
+    type: 'Hash160',
+    value: param,
+  }),
+  uInt256: (param) => ({
+    type: 'Hash256',
+    value: param,
+  }),
+  ecPoint: (param) => ({
+    type: 'ByteArray',
+    value: param,
+  }),
+  number: (param) => ({
+    type: 'Integer',
+    value: param,
+  }),
+  string: (param) => ({
+    type: 'String',
+    value: param,
+  }),
+  boolean: (param) => ({
+    type: 'Boolean',
+    value: param,
+  }),
+  buffer: (param) => ({
+    type: 'ByteArray',
+    value: param,
+  }),
+  object: (param) => ({
+    type: 'Array',
+    value: Object.entries(param).map(([k, v]) => [paramToArgDataType(k), paramToArgDataType(v)]),
+  }),
+};
+
+const paramToArgDataType = (param: ScriptBuilderParam | undefined): Argument =>
+  common.paramTo<Argument>(param, paramToArgDataTypeCallbacks);
+
+export class DapiUserAccountProvider<TProvider extends Provider> extends UserAccountProviderBase<TProvider>
+  implements UserAccountProvider {
+  public readonly currentUserAccount$: Observable<UserAccount | undefined>;
+  public readonly userAccounts$: Observable<ReadonlyArray<UserAccount>>;
+  public readonly networks$: Observable<ReadonlyArray<NetworkType>>;
+  private readonly dapi: Dapi;
+  private mutableInitPromise: Promise<void>;
+  private readonly currentAccountInternal$: BehaviorSubject<UserAccount | undefined>;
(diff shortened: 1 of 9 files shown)


packages/neo-one-client-core/src/user/DapiUserAccountProvider.ts
```diff
@@ -0,0 +1,335 @@
...
+    super({ provider });
+    this.dapi = dapi;
+    this.initPromise = this.init();
+
+    this.userAccountsInternal$ = new BehaviorSubject<ReadonlyArray<UserAccount>>([]);
+    this.currentAccountInternal$ = new BehaviorSubject<UserAccount | undefined>(undefined);
+    this.networksInternal$ = new BehaviorSubject<ReadonlyArray<NetworkType>>([]);
+    this.defaultNetworkInternal$ = new BehaviorSubject<NetworkType | undefined>(undefined);
+
+    this.dapi.addEventListener('ACCOUNT_CHANGED', ({ address, label }) => {
```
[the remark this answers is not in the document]
afragapane: That is correct.  https://github.com/neo-project/proposals/blob/bc18a72693185c6be2d6b8575d7cfc7850336dab/nep-dapi.mediawiki#ACCOUNT_CHANGED

They have an `Account` object which stores an address and a label and a `PublicKey` object which stores an address and a public key.  This listener passes in an `Account` object.

packages/neo-one-client-core/src/user/DapiUserAccountProvider.ts
```diff
@@ -0,0 +1,394 @@
...
+
+      return {
+        ...argBase,
+        txAttrUsage: attribute.usage,
+      };
+    });
+  }
+
+  private async init(): Promise<void> {
+    await new Promise<void>((resolve) => {
```
dicarlo2: This needs a timeout, otherwise the application might just hang forever when they call a method that depends on this.

packages/neo-one-client-core/src/user/DapiUserAccountProvider.ts
```diff
@@ -0,0 +1,394 @@
...
+          ).catch(() => {
+            // do nothing
+          });
+        }
+
+        firstCall = undefined;
+      });
+    };
+
+    setAccountChangedEventListener();
```
dicarlo2: I think this event listener part can be simplified to just

```
    const updateAccount = (network: string, dapiAccount: DapiAccount) =>
      this.dapi.getPublicKey().then(({ publicKey }) => {
        this.updateUserAccount({
          network,
          publicKey,
          ...dapiAccount,
        });
      }).catch((err) => {
        if (onError) {
          onError(err);
        }
      }); 

    let network: string | undefined;
    let account: DapiAccount | undefined;
    this.dapi.addEventListener(ACCOUNT_CHANGED, (dapiAccount) => {
      account = dapiAccount;
      if (network !== undefined) {
        updateAccount(network, dapiAccount);
      }
    });

    this.dapi.addEventListener(NETWORK_CHANGED, (dapiNetworks) => {
      network = dapiNetworks.defaultNetwork;
      this.updateNetworks(dapiNetworks);
      if (account !== undefined) {
        updateAccount(dapiNetworks.defaultNetwork, account);
      }
    });   
```

dicarlo2: Actually, the `getPublicKey` should also check if the address matches, i.e. it should look something like

```
    const updateAccount = (network: string, dapiAccount: DapiAccount) =>
      this.dapi.getPublicKey().then(({ address, publicKey }) => {
        if (address === dapiAccount.address) {
          this.updateUserAccount({
            network,
            publicKey,
            ...dapiAccount,
          });
        }
      }).catch((err) => {
        if (onError) {
          onError(err);
        }
      });   
```

Otherwise it's possible that `getPublicKey` returns a key for a different address from the account that's passed in.

packages/neo-one-client-core/src/user/DapiUserAccountProvider.ts
```diff
@@ -0,0 +1,394 @@
...
+      const argBase = paramToArgDataType(attribute.data);
+
+      return {
+        ...argBase,
+        txAttrUsage: attribute.usage,
+      };
+    });
+  }
+
+  private async init(): Promise<void> {
```
dicarlo2: Now that I think about it, this pattern has some issues when used here. If any of these fail, then `DapiUserAccountProvider` will permanently throw errors whenever a method is called that awaits the init promise. Instead what it should do is everywhere that we `await this.initPromise` should catch the error, reattempt initialization by setting `this.initPromise = this.init()` as we do in the constructor, and then throw the error. That way subsequent calls have their own attempts at initialization, but we only ever initialize _successfully_ once.

packages/neo-one-client-core/src/user/DapiUserAccountProvider.ts
```diff
@@ -0,0 +1,394 @@
...
+import {
+  ExecuteInvokeClaimOptions,
+  ExecuteInvokeMethodOptions,
+  ExecuteInvokeScriptOptions,
+  Provider,
+  UserAccountProviderBase,
+} from './UserAccountProviderBase';
+
+// Maintain consistency with @neo-one/client-common/src/ScriptBuilder.ts
+const paramToArgDataType = (value: ScriptBuilderParam | undefined): Argument => {
```
dicarlo2: I wonder if there's a way that we can consolidate the logic between here and `ScriptBuilder` as opposed to the comment, which may be easily missed if we change something small in the function (here or there).

packages/neo-one-client-core/src/user/DapiUserAccountProvider.ts
```diff
@@ -0,0 +1,360 @@
...
+          });
+        }
+      });
+
+    let network: string | undefined;
+    let account: DapiAccount | undefined;
+    this.dapi.addEventListener(ACCOUNT_CHANGED, (dapiAccount) => {
+      account = dapiAccount;
+      if (network !== undefined) {
+        updateAccount(network, dapiAccount).catch((err) => {
```
dicarlo2: Move the `catch` to inside of `updateAccount`

packages/neo-one-client-core/src/user/DapiUserAccountProvider.ts
```diff
@@ -0,0 +1,360 @@
...
+          }
+        });
+      }
+    });
+
+    this.dapi.addEventListener(NETWORK_CHANGED, (dapiNetworks) => {
+      network = dapiNetworks.defaultNetwork;
+      this.updateNetworks(dapiNetworks);
+      if (account !== undefined) {
+        updateAccount(dapiNetworks.defaultNetwork, account).catch((err) => {
```
dicarlo2: Same here

packages/neo-one-client-core/src/user/DapiUserAccountProvider.ts
```diff
@@ -0,0 +1,360 @@
...
+    return this.networksInternal$.getValue();
+  }
+
+  protected async executeTransfer(
+    transfers: ReadonlyArray<Transfer>,
+    from: UserAccountID,
+    _attributes: readonly Attribute[],
+    networkFee: BigNumber,
+  ): Promise<TransactionResult<TransactionReceipt, InvocationTransaction | ContractTransaction>> {
+    try {
```
dicarlo2: Extract this bit of logic into a method so it's not duplicated everywhere

packages/neo-one-client-core/src/user/DapiUserAccountProvider.ts
```diff
@@ -0,0 +1,360 @@
...
+        ...argBase,
+        txAttrUsage: attribute.usage,
+      };
+    });
+  }
+
+  private async init(): Promise<void> {
+    await new Promise<void>(async (resolve, reject) => {
+      this.dapi.addEventListener('READY', resolve);
+      setTimeout(reject, 10000);
```
dicarlo2: Let's reject with an error that indicates we hit a timeout.

PR #1690, opened 2019-08-21
feat(e2e): node-bin e2e consensus test

danwbyrne:
add consensus test to `@neo-one/node-bin`.

Also tweak `@neo-one/logger` so that child log levels get set by `setGlobalLogLevels`.


packages/neo-one-node-bin/src/__data__/fetchBlockCount.ts
```diff
@@ -0,0 +1,24 @@
+import fetch from 'cross-fetch';
+
+export const fetchBlockCount = async (endpoint: string) => {
```
dicarlo2: Remove and just use a `NEOONEDataProvider` to fetch the block count.

dicarlo2:
Just need to fix the lint error.

Issue #1684, opened 2019-08-20
Add switcheo exchange e2e test


dicarlo2:
### Description

Add `switcheo` project with a typescript version of their exchange for testing.

afragapane:
Closed by #1766

PR #1707, opened 2019-08-26
test(exchange): add exchange contract based on Switcheo for testing

afragapane:
### Description of the Change
Added an exchange contract based on Switcheo's exchange and some tests.  In this PR, the contract allows the exchange of NEP5 tokens without accounting for Fees.

Things to Add after this PR:
- Allow exchange of system assets
- Account for Fees

It seems that the contract becomes too large if I try to add in much more.  I think it may be hitting the maximum size limit for contracts when more code is added.  

### Applicable Issues
#1684


afragapane:
Moved to: https://github.com/neo-one-suite/neo-one/pull/1766

Issue #1725, opened 2019-08-31
Test javascript codegen in e2e test


dicarlo2:
### Description

Needs an e2e test that at least verifies the codegen is valid javascript code. (language = 'javascript' in configuration)

afragapane:
Closed by: https://github.com/neo-one-suite/neo-one/pull/1731

Issue #1724, opened 2019-08-31
Test angular and vue codegen in e2e test


dicarlo2:
### Description

The angular and vue codegen needs a test that at least verifies that it emits valid javascript and typescript code. (framework = 'angular' and framework = 'vue')

afragapane:
Closed by: https://github.com/neo-one-suite/neo-one/pull/1731

PR #1731, opened 2019-09-03
test(e2e codegen): add e2e tests to verify codegen for javascript, angular, and vue

afragapane:
### Description of the Change
Add e2e tests to verify that codegen works properly for both typescript and javascript for the react, angular, and vue frameworks.  

### Test Plan
Verify tests pass and that the code produced is valid javascript/typescript.  Will also be testing that code produced by angular and vue work as front-ends.  

### Applicable Issues
#1724, #1725


packages/neo-one-cli/package.json
```diff
@@ -36,6 +36,7 @@
     "yargs": "^14.0.0"
   },
   "devDependencies": {
+    "@angular/core": "8.2.4",
```
dicarlo2: Needs a caret

PR #1732, opened 2019-09-03
fix(web): fix developer tools clipping

### LINKED ISSUE #1726: Fix hidden developer tools
### Description

When the developer tools are expanded, the iframe container should stay open at least as large as the expanded developer tools. Currently the iframe container resizes down to the small square size when you mouse away from the developer tools bar.

### Steps to Reproduce

Tested/seen in the neo-one playground, but I assume it's similar behavior on the courses.

1. Start the playground
2. Click on one of the playground menu items (e.g. Escrow)
3. Click on the developer tools icon in the lower left
4. Move mouse away and see that it resizes the container, showing only a small box of the bar.

**Expected behavior:** The container stays open

**Actual behavior:** It does not

### Versions

Master

danwbyrne:
Fix the developer tools clipping. Also add an `ifProp` switch to stop some initial transition clipping, also changes the behavior of the size observer. Works much better now after test and looking at the i-frame in different instances.

Fixes #1726

### DIFF
--- a/packages/neo-one-developer-tools-frame/src/ResizeHandler.ts
+++ b/packages/neo-one-developer-tools-frame/src/ResizeHandler.ts
@@ -67,7 +67,7 @@ export class ResizeHandler {
       this.mutableResizeTimeout = undefined;
       this.resize();
       // tslint:disable-next-line no-any
-    }, 1500) as any;
+    }, 1000) as any;
   };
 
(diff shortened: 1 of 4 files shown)


dicarlo2:
lgtm, just need to update the snapshots for the codegen package.

PR #1733, opened 2019-09-04
fix(tests): disable logs in withContract tests

### LINKED ISSUE #1727: Disable logging in `withContracts` tests
### Description

Title.

### Steps to Reproduce

1. Clone neo-one-playground
2. `yarn test`

**Expected behavior:** Clean passing tests

**Actual behavior:** Passing tests with a ton of log spam

danwbyrne:
title.

Fixes #1727


packages/neo-one-smart-contract-test-common/src/withContracts.ts
```diff
@@ -105,8 +113,18 @@ export const withContracts = async <T>(
...
+  {
+    ignoreWarnings = false,
+    deploy = true,
+    autoConsensus = true,
+    autoSystemFee = true,
+    logging = false,
+  }: WithContractsOptions = {},
 ): Promise<void> => {
+  if (!logging) {
+    disableConsoleLogForTest();
```
dicarlo2: I think we want to leave these enabled - otherwise users don't have a good way to debug using `console.log` in their smart contracts.

PR #1830, opened 2019-10-22
docs(deploy): update docs on deployment

afragapane:
### Description of the Change

Added some rough docs on deployment.  Kept intentionally vague until we work with users on the feature.


spencercorwin:
LGTM

PR #1921, opened 2019-12-09
P2p models & Address Script Construction

### LINKED ISSUE #1878: [Neo 3.0] Update Address Script Construction
https://github.com/Tommo-L/NEO3-Development-Guide#wallet

Single and multi-signature addresses have had their checksigs altered slightly, doesn't seem like a very aggressive change and should be easy to implement.

danwbyrne:
Check-in point for model updates. Block/Transaction changes have been made to the `types` files of `@neo-one/client-common` and several of the errors implicated by this have been addressed.

Also the functions that handle constructing address scripts has been updated. This should close #1878 but I'll leave it open for now.

A lot more is still broken but I think this is a good check point so that we don't work over each other.

### DIFF
--- a/packages/neo-one-node-core/src/Block.ts
+++ b/packages/neo-one-node-core/src/Block.ts
@@ -14,6 +14,7 @@ import { BN } from 'bn.js';
 import { Account, AccountKey } from './Account';
 import { Asset, AssetKey } from './Asset';
 import { BlockBase } from './BlockBase';
+import { ConsensusData } from './ConsensusData';
 import { MerkleTree } from './crypto';
 import { VerifyError } from './errors';
 import { Header, HeaderKey } from './Header';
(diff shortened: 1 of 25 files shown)


danwbyrne:
I left some `TODOs` in but I think we should come back to them / get multiple eyes on the code. Mostly around functions that read bytes directly or calculate size differently than they used to.

danwbyrne:
Also not sure how `rush change` is going to work in the long run if we don't coalesce the changelogs often. Hopefully if we mark several things as "major" changes it won't bump it for each one. i.e, 3 major changes should still only apply 1 major bump, as opposed to going from 2.0.0 -> 5.0.0

danwbyrne:
Need to add tests

packages/neo-one-node-core/src/Block.ts
```diff
@@ -44,11 +45,11 @@ export interface BlockAdd {
   readonly version?: number;
   readonly previousHash: UInt256;
   readonly merkleRoot?: UInt256;
-  readonly timestamp: number;
+  readonly timestamp: BN;
```
spencercorwin: Are we changing this for general integer safety?  The change to BN from number is causing problems elsewhere when the timestamp needs to be added to another number that isn't a BN. It's likely not a big deal to deal convert a number to BN when needing to use it together with the timestamp, but I want to understand why we might need timestamp to be a BN

Issue #1876, opened 2019-12-02
Contract Manifest - Node


afragapane:
### Description
- https://github.com/neo-project/neo/blob/master/src/neo/SmartContract/Manifest/ContractManifest.cs
- Add models for Manifest, Group, Permissions, ect. in `@neo-one/client-full-common`
- Update Contract to use the updated Models in `@neo-one/node-core`
- Update ContractStackitem to use the updated Contract in `@neo-one/node-vm`
- Update syscalls for `Create`, and `Migrate`/`Update` to use the manifest
- Add tests for all of the above

PR #1925, opened 2019-12-12
refactor(contracts3.0): Add NEO 3.0 contracts to the vm

afragapane:
### Description of the Change
Adds serializable/deserializable Contract, ContractManifest, ContractPermissions, ContractPermissionDescriptors, ContractABI, ContractEvents, ContractParameterDeclarations, and ContractGroups. 

### Test Plan
Added unit tests that confirm serialize/deserialize works and json works.  

### Benefits
Lays the ground work for serializable/deserializable contracts using the new Manifest style.  Probably will need some tweaking, but getting this base in will allow other work to progress. 

### Possible Drawbacks
Note: This does not yet include the "Wildcard" type in the neo-project.  Unsure if we need this yet, but worth keeping an eye on.

These are also likely missing necessary methods such as "CanCall" on the contract.  We should add these as needed in the appropriate locations.

### Applicable Issues
#1876


packages/neo-one-node-core/src/contract/abi/ContractEvent.ts
```diff
@@ -0,0 +1,73 @@
...
+export const sizeOfContractEvent = ({
+  name,
+  parameterDeclarations,
+}: {
+  readonly name: string;
+  readonly parameterDeclarations: readonly ContractParameterDeclaration[];
+}) => IOHelper.sizeOfVarString(name) + IOHelper.sizeOfArray(parameterDeclarations, (parameter) => parameter.size);
+
+export const deserializeContractEventWireBase = ({ reader, context }: DeserializeWireBaseOptions): ContractEvent => {
+  const name = reader.readVarString(252);
```
danwbyrne: I feel like I've had to read `252` bytes off of something somewhere as well. Maybe we should make this a constant on `utils`? Or even have a reader function for doing 252 (just to add some context to *why* we are reading 252 bytes)

packages/neo-one-node-core/src/contract/abi/ContractFunction.ts
```diff
@@ -0,0 +1,91 @@
...
+  IOHelper.sizeOfVarString(name) +
+  IOHelper.sizeOfArray(parameterDeclarations, (parameter) => parameter.size) +
+  // returnType
+  IOHelper.sizeOfUInt8;
+
+export const deserializeContractFunctionWireBase = ({
+  reader,
+  context,
+}: DeserializeWireBaseOptions): ContractFunction => {
+  const name = reader.readVarString(252);
```
danwbyrne: here as well

packages/neo-one-node-core/src/contract/abi/parameters/ContractParameterDeclaration.ts
```diff
@@ -0,0 +1,70 @@
...
+
+export const sizeOfContractParameterDeclaration = ({ name }: { readonly name: string }) =>
+  IOHelper.sizeOfVarString(name) + IOHelper.sizeOfUInt8;
+
+export const deserializeContractParameterDeclarationWireBase = ({
+  reader,
+}: DeserializeWireBaseOptions): ContractParameterDeclaration => {
+  const type = assertContractParameterType(reader.readUInt8());
+  // throw new Error(reader.remainingBuffer.toString('hex'));
+  const name = reader.readVarString(252);
```
danwbyrne: here too :)

PR #1961, opened 2020-01-20
refactor(syscall tests): Split syscall tests to be less clutered

afragapane:
### Description of the Change
Split syscall tests into multiple files.  The file was getting cumbersome to work with, this should help make things easier.  

### Test Plan
Confirmed tests that should still pass, pass.


packages/neo-one-node-vm/src/__data__/syscallRunner.ts
```diff
@@ -0,0 +1,272 @@
...
+      }
+    }
+
+    return result;
+  }
+
+  return value;
+};
+
+export const syscallRunner = (sysCalls: readonly TestCase[]) => {
```
spencercorwin: Rename to `runSyscalls()`

packages/neo-one-node-vm/src/__tests__/syscalls/binary.test.ts
```diff
@@ -0,0 +1,207 @@
...
+            .writeUInt8(StackItemType.ByteArray)
+            .writeVarBytesLE(Buffer.from('value', 'utf8'))
+            .toBuffer(),
+        )
+        .toBuffer(),
+    ],
+
+    gas: FEES[500_000],
+  },
+] as readonly TestCase[];
```
spencercorwin: Just noticed that we've been type casting this array, which seems less safe than tagging the array as `SYSCALLS: readonly TestCase[]`

Issue #1874, opened 2019-12-02
Neo 3.0


afragapane:
### Description
All work required for neo-one to be NEO 3.0 ready.

PR #1969, opened 2020-01-31
fix(syscalls): update system.contract.create and system.contract.create

spencercorwin:
### Description of the Change

- Updates SysCalls `System.Contract.Create` and `System.Contract.Update`
- Updates/ creates unit tests for these two SysCalls
- Deletes any SysCalls that no longer exist
- Adds `fromJSON()` static methods to all Manifest models

### Test Plan

- `rush test -t packages/neo-one-client-common/src/__tests__/common.test.ts`
- `rush test -t packages/neo-one-node-vm/src/__tests__/syscalls/crypto.test.ts`
- `rush test -t packages/neo-one-node-vm/src/__tests__/syscalls/contract.test.ts`

### Alternate Designs

None.

### Benefits

Closer to NEO 3.0.

### Possible Drawbacks

The NF team will likely change these in the future, so these shouldn't be considered done.

Note that changes need to be made to other models, like the `Contract` model, which will change this PR after it's merged. My next work will be to update the `Contract` with its new `ID` property.

### Applicable Issues

#1874


packages/neo-one-node-vm/src/__data__/transactions.ts
```diff
@@ -86,6 +134,7 @@ export const conciergeContract = createContract({
   manifest: createContractManifest({
     features: ContractPropertyState.HasStoragePayable,
     abi: createContractAbi({
+      hash: crypto.hash160(conciergeContractScript),
```
afragapane: nit: use `toScriptHash`

packages/neo-one-node-vm/src/syscalls.ts
```diff
@@ -1120,58 +1127,111 @@ export const SYSCALLS: { readonly [K in SysCallEnum]: CreateSysCall } = {
...
+              safeMethods: contract.manifest.safeMethods,
+              extra: contract.manifest.extra,
+            }),
+          });
+
+          await context.blockchain.contract.add(contract);
+
+          if (contract.manifest.hasStorage) {
+            await context.blockchain.storageItem
+              .getAll$({ hash: context.scriptHash })
```
afragapane: Change this to check if any storage items exist.  Maybe use a pattern similar to `Neo.Storage.Find` with `toArray` or make sure of the `count` operator: https://rxjs-dev.firebaseapp.com/api/operators/count.

packages/neo-one-client-common/src/common.ts
```diff
@@ -51,6 +51,30 @@ const uInt160ToString = (value: UInt160 | UInt160Hex): string =>
...
+
+  if (isUInt160(tryUInt160)) {
+    return tryUInt160;
+  }
+  if (isECPoint(tryECPoint)) {
+    return tryECPoint;
+  }
+
+  return undefined;
+};
```
spencercorwin: @afragapane This is one more function I created that I want to make sure makes sense. It's used when converting a manifest string to a Manifest instance. I created this for the `hashOrGroup` attribute in the `ContractPermissionDescriptor` object. The string could represent a `UInt160` or could be an `ECPoint`. This should work, and I wrote unit tests for it, but I just want another set of eyes on it.

afragapane: What you have works, but I think it could be nice to simplify a little bit by returning inside of the try blocks since they already check that the return is a valid UInt160/ECPoint.  Something like:
```
try {
  const tryUint160 = stringToUInt160(stringIn);
  return tryUint160;
}
```

PR #1972, opened 2020-02-05
fix(codecoverage): fixup codecoverage post rush + misc updates

danwbyrne:
- unit tests now use source code again instead of built code.
- code coverage should be working again with that change.
- add better logging to the chainfile sync


packages/neo-one-build-tests/jest/base.js
```diff
@@ -32,7 +32,7 @@ module.exports = ({ path }) => ({
     `${jest_root}/serializers/buffer.js`,
   ],
   testPathIgnorePatterns,
-  transformIgnorePatterns: ['node_modules', '/packages/.*/lib/'],
+  transformIgnorePatterns: ['node_modules'],
```
spencercorwin: Is there a reason we don't need to include `lib` in this? I'm sure there is, I'm just curious

packages/neo-one-build-tests/jest/e2e.js
```diff
@@ -6,11 +6,7 @@ module.exports = {
   testRunner:
     '<rootDir>/packages/neo-one-build-tests/node_modules/jest-circus/runner',
   testRegex: '^.*/__e2e__/.*\\.test\\.tsx?$',
-  moduleNameMapper: {
-    '^@neo-one/ec-key': '@neo-one/ec-key',
-    '^@neo-one/boa': '@neo-one/boa',
-    '^@neo-one/csharp': '@neo-one/csharp',
-  },
+  moduleNameMapper: undefined,
```
spencercorwin: I don't think I've ever touched `e2e` tests, so curious again, but why remove the `moduleNameMapper` here?

packages/neo-one-build-tests/jest/unit.js
```diff
@@ -3,23 +3,14 @@ const base = require('./base');
 module.exports = {
   ...base({ path: 'test' }),
   displayName: 'unit',
+  moduleNameMapper: {
```
spencercorwin: I'm sure this relates to the answer to my comment on `e2e.js` above, but curious why we didn't have this here before?

packages/neo-one-build-tests/tsconfig.jest.json
```diff
@@ -27,10 +27,5 @@
       "bn.js": ["node_modules/@neo-one/build-tools/types/bn.js.d.ts"]
     }
   },
-  "exclude": [
-    "./packages/*/node_modules",
-    "./packages/*/lib",
-    "./packages/*/src/__data__/contracts/**/*.ts",
-    "./packages/*/template/**/*.ts?(x)"
-  ]
+  "exclude": ["./packages/*/src/__data__/contracts/**/*.ts", "./packages/*/template/**/*.ts?(x)"]
```
spencercorwin: Curious why we wouldn't still need to exclude `lib` and subdirectory `node_modules`? It looks like tsconfig only excludes the root-level `node_modules` by default

packages/neo-one-build-tools/src/builds/buildTypescript.ts
```diff
@@ -37,7 +37,6 @@ export const buildTypescript = (format: Format, pkgName?: string) => (
     format,
     gulp
       .src(glob)
-      .pipe(gulpPlumber())
```
spencercorwin: Remove `gulpPlumber` import above

Issue #1889, opened 2019-12-03
Upgrade to TS v3.8


spencercorwin:
TS v3.7 is out and need to be incorporated. Some API changes are breaking changes for us and need to be considered: https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#37-api-changes

PR #1980, opened 2020-02-19
feat(ts): upgrade TS to v3.8.1-rc

spencercorwin:
### Description of the Change

Upgrades TS to v3.8.1-rc. Makes a few changes to make TS v3.8.1-rc compatible with our stuff. Additions to our compiler for supporting new language features will come in another PR.

I split this into two commits for a good reason. The TS upgrade apparently wreaks havoc on the types for `@emotion`, which is used all throughout these packages: `neo-one-developer-tools-frame`, `neo-one-editor`, `neo-one-react-core`, `neo-one-react-common`, and `neo-one-website`. This forced me to go through and change A LOT of generics, many of which I seemingly couldn't do anything besides pass in an `any` type. It looks like this problem has been recognized and fixed in the `@emotion` GitHub. See [here](https://github.com/emotion-js/emotion/pull/1501). But it looks like that won't be deployed until the release of Emotion v11. See [here](https://github.com/emotion-js/emotion/milestone/1). So ideally they release that fix in the near future, then we can upgrade, then easily revert all these weird generics/`any` types and add back the lost type-safety in these packages using `git revert`

### Test Plan
- `rush test`
- `rush e2e`

### Alternate Designs

As a note, there was a breaking change in their code that caused us to have to change some code in `symbol.ts -> getAliasedSymbol()`. It's very difficult to tell where their changed happened that caused this breaking change, as they don't mention anything related to it, and none of the related `TypeChecker` methods or `Type` properties changed. I believe it's the original construction of the `aliasSymbol` property that caused this breaking change for us. See the commit [here](https://github.com/microsoft/TypeScript/commit/fd8f9904490ec7cf93e1010604e7a4bec2a56756). Mostly just for future note.

### Benefits

- We get to use the latest TS features in our own code from here on.

### Possible Drawbacks

- Prettier doesn't support TS v3.8 yet. They have an issue open for this [here](https://github.com/prettier/prettier/issues/7263). They are waiting for TS 3.8 support in `typescript-eslint` fo the new AST [here](https://github.com/typescript-eslint/typescript-eslint/pull/1465). It only affects one place in our code, so I think it's easy to ignore for now.
- Possibly unforeseen errors

### Applicable Issues

#1889

### DIFF
--- a/common/changes/@neo-one/cli-common-node/upgrade-ts_2020-02-26-01-16.json
+++ b/common/changes/@neo-one/cli-common-node/upgrade-ts_2020-02-26-01-16.json
@@ -1,0 +1,11 @@
+{
+  "changes": [
+    {
+      "packageName": "@neo-one/cli-common-node",
+      "comment": "Upgrade TS to v3.8.1-rc",
+      "type": "minor"
+    }
+  ],
+  "packageName": "@neo-one/cli-common-node",
+  "email": "spencercorwin@icloud.com"
+}
--- a/common/changes/@neo-one/cli/upgrade-ts_2020-02-26-01-16.json
+++ b/common/changes/@neo-one/cli/upgrade-ts_2020-02-26-01-16.json
@@ -1,0 +1,11 @@
+{
+  "changes": [
+    {
+      "packageName": "@neo-one/cli",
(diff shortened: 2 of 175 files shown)


spencercorwin:
The failing `nit` check is because of Prettier, which is acceptable.

PR #2005, opened 2020-04-13
feat(compiler): add support for outputting compiled contracts

danwbyrne:
### Description of the Change

Adds support for outputting compiled contract information to file. The new cli command is `compile` and has 2 flags `--path` the directory of source contracts and `--outDir` to output.

### Test Plan

You can test this by running
```
rush build
node packages/neo-one-cli/bin/neo-one compile --path <testContractDir> --outDir ./testDir
```
there are example contract projects in the `__data__` folder of `neo-one-cli`.

### Alternate Designs
Right now we just output the result of our compilation to a `.json` file but we could change the format of the output depending on what the community needs.

### Benefits

Should make the typescript compiler more usable for third parties that don't require the entire `neo-one` suite.

### DIFF
--- a/common/changes/@neo-one/cli/compiler_2020-04-13-22-53.json
+++ b/common/changes/@neo-one/cli/compiler_2020-04-13-22-53.json
@@ -1,0 +1,11 @@
+{
+  "changes": [
+    {
+      "packageName": "@neo-one/cli",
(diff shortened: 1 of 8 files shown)


spencercorwin:
Nice!

PR #2021, opened 2020-05-01
chore(deps): update caniuse-lite dependency

danwbyrne:
title, will affect a few other dependencies too since I just ran `rush update --full` but should be fine.

Some changes in a new @types/pino version don't seem right, so I've just ignored them for now and we can try to update later. When I look at `pino/index.js` the functions which the types warn about not being there are definitely there.

### DIFF
--- a/common/config/rush/yarn.lock
+++ b/common/config/rush/yarn.lock
@@ -10,9 +10,9 @@
     tslib "^1.9.0"
 
 "@babel/cli@^7.5.5":
-  version "7.7.4"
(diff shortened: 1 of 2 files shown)


spencercorwin:
I looked into the failing checks here. The Cypress check is failing due to something else. Then there's 3 unit tests failing. Two are snapshots that can be updated. The third is a test that is failing due to Frag's upgrades to the `DataProvider` and that will be fixed as we finish integrating the C# node. Then there's the obvious linter check on `nodePath` which we can tell `tslint` to ignore that line.

PR #2023, opened 2020-05-04
fix(docs): fix slight mismatch in deploy doc

spencercorwin:
### Description of the Change

Go to https://neo-one.io/docs/deployment and see the "Upgrade" section. We _say_ that the `approveUpgrade` method has to be a `protected` method. But in the code example the `approveUpgrade` method is a `public` method. I think this should say `public` method so as to avoid confusion.

Someone should make sure I'm correct here before merging.


danwbyrne:
"someone"

Issue #1759, opened 2019-09-10
Add documentation on deploying


dicarlo2:
### Description

Title

spencercorwin:
https://neo-one.io/docs/deployment

PR #2052, opened 2020-05-28
feat(docs): Add initial documention for deploying to a public network

danwbyrne:
Adds documentation for how to deploy a smart contract to the test/main net

#1759


packages/neo-one-website/docs/2-advanced-guides/09-configuration-options.md
```diff
@@ -62,3 +62,29 @@ export default {
...
+    })
+  }
+}
+```
+
+::: warning
+
+Note
+
+while hard coding the `LocalUserAccountProvider` is a viable option in testing this also requires storing a `privateKey` as plain text in a file that would traditionally be checked into version control, i.e github. For this reason we recommend only using a hard coded value for local on-the-fly testing / debugging.
```
spencercorwin, line 88: Tiny nit: `While hard coding...`
Capitalize the `while`

Issue #2025, opened 2020-05-04
Add e2e test for new command `neo-one compile`


spencercorwin:
### Description

Just need to make sure to add a new e2e test for `neo-one compile` CLI command which was added in #2005

PR #2064, opened 2020-06-16
feat(e2e): add e2e test for neo-one compile

spencercorwin:
### Description of the Change

Adds `outDir` to the `neo-one` config file for specifying where to put the outputted ABI files from the `neo-one compile` command which was added earlier. The command still works the  same except that if the command is run without `--outDir` specified then it will use the `outDir` path specified in the `.neo-one.config.js` file, which defaults to `neo-one/compiled`.

This also adds an E2E test for the `neo-one compile` command.

The unit test `blockchain.test.ts` is failing locally, but apparently not in CI. Not sure why.

### Test Plan

`rush e2e -t packages/neo-one-cli/src/__e2e__/cmd/compile.test.ts`

### Alternate Designs

None.

### Benefits

Test coverage for compile command.

### Possible Drawbacks

None.

### Applicable Issues

#2025


packages/neo-one-cli/src/cmd/deploy.ts
```diff
@@ -6,7 +6,7 @@ import { deploy } from '../deploy';
 export const command = 'deploy';
 export const describe = 'Deploys the project using the migration file.';
 export const builder = (yargsBuilder: typeof yargs) =>
-  yargsBuilder.string('network').describe('network', 'Network to run the migration on.').default('network', 'test');
```
danwbyrne, line 9: this should still be 'test'; see https://github.com/neo-one-suite/neo-one/blob/1fe420d7a03f1d32e87a4c56a6871d22a2e2d240/packages/neo-one-cli-common-node/src/networks.ts#L26

spencercorwin: Ahh. Got it.

Issue #2061, opened 2020-06-09
Keep NEO Tracker updated in NEO•ONE


spencercorwin:
### Description

We want to have NEO•ONE always use the latest version of NT. To do that we think it would be best to make the config turn off NT by default. We'd then add instructions/docs/config comments to tell the user that they can use NT but they'd need to install it as a dependency in their project. That way they're always using the latest version of NT.

spencercorwin:
If NT is enabled (skip: false) then in a try/catch statement try to start NT. If it's not found then assume that NT isn't installed on the user's machine or in the user's project. Then just log a message that says "@neotracker/core not found. Try installing @neotracker/core with `yarn add @neotracker/core`". Then they install a version of NT themselves.

PR #2066, opened 2020-06-17
feat(nt): make neotracker dev dependency to decouple

spencercorwin:
### Description of the Change

Removes `@neotracker/core` as a dependency in `@neo-one/cli`. This way we won't have to keep updating the `@neotracker/core` dependency. Instead it will be decoupled from neo-one. 

### Test Plan

Let the e2e tests run (and all other checks). They should all still pass, even the neotracker ones. I tested the failure by running uninstalling neotracker completely, then running the same e2e test and saw the expected error.

### Alternate Designs

Tried using `await import('@neotracker/core')` instead of testing the dependency with `require.resolve()` but ran into problems where it apparently couldn't find neotracker and would fail the build.

### Benefits

Less maintenance work, and always the latest version of NT for NEO•ONE users.

### Possible Drawbacks

Shifts the burden to NEO•ONE users to install `@neotracker/core` if they want that functionality.

### Applicable Issues

#2061


danwbyrne:
yep looks like exactly what we wanted :+1:

Issue #2073, opened 2020-06-29
Update NEO Tracker dependency


spencercorwin:
### Description

Release v1.4.0 of NEO Tracker and update dependency in NEO•ONE.

The neotracker-ico project related E2E tests are now failing after adding the RushJS post-install script to remove @neo-one packages from `common/temp/node_modules`. So we need to fix or find a workaround for that too.

PR #2100, opened 2020-07-16
fix(tests): update neotracker version for cli tests

danwbyrne:
### Description of the Change

Update neotracker dependency to 1.4.0 to get rid of any traces of opencensus

### Test Plan

```
rush e2e -t neotracker
```

### Alternate Designs

We tried a few things to essentially simulate neotracker being a part of this monorepo but everything hit a wall one way or another whether it be dependency resolution or problems rush encountered during its linking when I attempted to just create the correct `@neo-one` symlinks for neotracker myself.

### Benefits

opencensus should never creep up again

### Possible Drawbacks

we are going back to NOT deleting the `@neo-one` dependencies post-install, let neotracker do whatever its going to do in practice it should resolve to whatever version of `@neo-one` the user has installed (which could cause its own problems down the line since we don't explicitly test that interaction like we were trying to do)

### Applicable Issues

#2073


spencercorwin:
Needs `rush change`. And `rush install` is failing?

packages/neo-one-cli/src/__e2e__/cmd/start/neotracker.test.ts
```diff
@@ -18,7 +18,6 @@ describe('start network', () => {
 
     execAsync('start neotracker');
     // this allows us to run the e2e tests in parallel, successfully
```
spencercorwin, line 20: We should also remove this comment.

packages/neo-one-cli/src/__e2e__/cmd/stop/neotracker.test.ts
```diff
@@ -18,7 +18,6 @@ describe('start network', () => {
 
     execAsync('start neotracker');
     // this allows us to run the e2e tests in parallel, successfully
```
spencercorwin, line 20: We should also remove this comment.

PR #2137, opened 2020-08-13
feat(vm): initial support for C# VM

danwbyrne:
## Description

Adds the temporary package `@neo-one/csharp-vm` which will later just become the new `@neo-one/node-vm`.

## Known Issues
1. #2140 `Assembly.GetEntryAssembly()` returns null when running in our CLR instance. We'll need a solution to this as at the moment the only solution is to use ProjectReferences to the `Neo` module and using `Assembly.GetExecutingAssembly()` instead. See `lib/EngineDispatcher.csproj` and make appropriate adjustments for your machine. 

2. #2141 Setup can be slightly different for windows / osx / linux at the moment, and I am only confident in it working with OSX / windows for now. See the `packages/neo-one-csharp-vm/README.md` for basic setup instructions.

## Additional Changes

We also update several `@neo-one/node-core` Block models to correspond with neo3 changes / for use in testing the current progress of `@neo-one/csharp-vm`.


packages/neo-one-node-core/src/ConsensusData.ts
```diff
@@ -0,0 +1,64 @@
...
+  DeserializeWireBaseOptions,
+  SerializeWire,
+  createSerializeWire,
+} from './Serializable';
+import { ConsensusDataJSON, BinaryWriter, IOHelper, crypto, UInt256 } from '@neo-one/client-common';
+import { BN } from 'bn.js';
+import { utils } from './utils';
+
+// TODO: could put this somewhere else it is possible it changes?
+const defaultValidatorsCount = new BN(7);
```
spencercorwin, line 13: Put it in `./constants`? And rename `DEFAULT_VALIDATORS_COUNT`?

packages/neo-one-node-core/src/ConsensusData.ts
```diff
@@ -0,0 +1,64 @@
...
+
+export interface ConsensusDataAdd {
+  readonly primaryIndex: number;
+  readonly nonce: BN;
```
spencercorwin, line 17: I think `ConsensusData` needs an optional hash: `readonly hash?: UInt256`

packages/neo-one-node-core/src/ConsensusData.ts
```diff
@@ -0,0 +1,64 @@
...
+  public readonly nonce: BN;
+  public readonly serializeWire: SerializeWire = createSerializeWire(this.serializeWireBase.bind(this));
+  public readonly sizeInternal = utils.lazy(
+    () => IOHelper.sizeOfVarUIntLE(this.primaryIndex) + IOHelper.sizeOfUInt64LE,
+  );
+  public readonly hashInternal = utils.lazy(() => crypto.hash256(this.serializeWire()));
+
+  public constructor({ primaryIndex, nonce }: ConsensusDataAdd) {
+    this.primaryIndex = primaryIndex;
+    this.nonce = nonce;
```
spencercorwin, line 42: `this.hashInternal = hashIn === undefined ? utils.lazy(() => crypto.hash256(this.message)) : () => hashIn`   ?

See `ConsensusData.ts` here: https://github.com/neo-one-suite/neo-one/pull/2131/files#diff-eaa1ebfacb78fd619fd18a95571c3aa4

Issue #2060, opened 2020-06-08
Swap TS VM for C# VM


spencercorwin:
Packages:
- `node-vm`

This is a first tier priority.

Issue #2009, opened 2020-04-21
3.0 Release


danwbyrne:
* First level packages
    - [x] node-blockchain
    - [x] node-core
    - [x] node-vm
    - [x] node-storage-cache
        - Creates the storage cache
        - Not actually used anywhere
    - [x] node-storage-common
        - Exports storage helpers for storage keys in storage-levelup and storage-cache
    - [x] node-storage-levelup
        - 90% done
        - Actually creates all the storage needed for node
    - [x] client
        - Just exports
    - [ ] client-full
        - Just exports
    - [ ] client-core
		- Current WIP by Spencer
        - Client APIs
    - [x] client-common
        - ~~Native contracts~~
        - Crypto
    - [x] client-full-common
    - [ ] client-full-core
		- Current WIP by Spencer
        - Client APIs
* Second level packages
    - [x] build-common
    - [x] build-tests
    - [x] build-tools
    - [x] build-tools-web
    - [x] cli
    - [ ] cli-common
    - [ ] cli-common-node
    - [x] node-offline
        - See if chain file sync changed
    - [x] node-rpc-handler
        - 80% done. Blocked by blockchain storage changes and other utils
    - [x] node-consensus
        - Handles consensus for the node
    - [ ] node-protocol
        - ~Creates the “node” part of the FullNode~
        - ~Handles connecting to peer nodes and sending/receiving messages~
        - ~Handles persisting blocks to the blockchain~
        - ~Takes in a blockchain~
        - Handles mempool and transactions
    - [x] node-network
        - Network passed into “Node” from node-protocol
        - Handles network endpoints and connections for the node
    - [x] node-neo-settings
        - 80% done, need private network settings support + some settings might be set in stone that weren't previously.
        - Contains all the settings and setup configs needed for the various types of nodes/networks
* Third level packages
    - [x] node
        - Creates FullNode class
        - Very minor changes coming last
    - [x] node-bin
    - [ ] client-switch
* Packages requiring minor to no updates/checks last
    - [x] ~suite~
    - [x] node-http-rpc
        - Sets up RPC server, very basic
    - [x] node-tools
        - Syncs gcloud bucket to local file system for node data

PR #2146, opened 2020-08-25
feat(vm): organize vm dispatcher, add initial snapshot support

danwbyrne:
### Description

- Organize `@neo-one/csharp-vm` into a new dispatcher model that is more compartmentalized. 
- Updated unit tests and added the `SnapshotHandler` class for handling 2 internal snapshots, `main` and `clone`.
- updated `verify.ts` in `@neo-one/csharp-blockchain` which will be further implemented in an upcoming PR.
- updated `ByteStringStackItem` value type to be Buffer instead of string, added `asString()` function and removed `_buffer` property.


spencercorwin:
#2060 
#2009

Issue #1975, opened 2020-02-10
Add Contract IDs to Contract Model


spencercorwin:
### Description

NEO has added a contract ID property to Contract Model. See the discussion and PR here: https://github.com/neo-project/neo/pull/1400

Make sure to implement the Contract ID usage where appropriate, like in various SysCalls and OpCodes and StorageItems. See their PR for guidance.

spencercorwin:
Needs to be done in Client as well for 3.0

Issue #1883, opened 2019-12-03
Update block structure


spencercorwin:
### Description

- Remove block headers/ change the block structure to match C# code
- Add Trimmed Blocks
- Update SysCalls/OpCodes that references transactions within a block
- Specifically update `System.Blockchain.GetTransactionFromBlock`
- Remove `HeaderStackItem`
- Remove `UInt256StackItem`? Not used in any SysCalls currently
- Remove `InvalidGetHeaderArgumentsError`
- Remove `ContainerTooLargeError`?

Issue #1881, opened 2019-12-02
Delete Alternative Transaction Models


danwbyrne:
https://github.com/Tommo-L/NEO3-Development-Guide#transactions

Transactions have been fundamentally reworked beginning with the removal of all transaction types besides a singular `transaction` model. This will probably be the most involved changes of the 3.0 client migration and may warrant other issues in the future. For now the tasks needed to be complete are:

- [x] Remove every transaction except `InvocationTransaction` and rename to just `Transaction`.
- [x] Update the current `Transaction` model to match the model changes made in 3.0.
- [ ] discard the UTXO model for NEO and GAS and use the account model from the native contract. We have a lot of code involving inputs/outputs and calculating balances so this will be ~80% removal of old code.

spencercorwin:
Covered by #2142 and #1975, etc.

PR #2147, opened 2020-08-27
feat(vm/blockchain): Add/Update more core types for the blockchain package

danwbyrne:
### Description

- Adds new / updated models for contracts
- Adds new / updated models for storage
- Adds interfaces for blockchain, settings, and storage


spencercorwin:
#2009 
#1975 
#2060 
#1883 
#1881

packages/neo-one-client-full-common/src/models/BaseState.ts
```diff
@@ -1,8 +0,0 @@
-export class BaseState {
-  public static readonly VERSION = 0;
-  public readonly version: number;
-
-  public constructor({ version: versionIn }: { readonly version?: number }) {
-    this.version = versionIn === undefined ? (this.constructor as typeof BaseState).VERSION : versionIn;
-  }
-}
```
spencercorwin, line 8: We use this in node-core. In `Account`, `Asset`, `TransactionData`, `Validator`, and `ValidatorsCount`. You may have already covered this in the `csharp`, if so then ignore. It might also make sense to just move this to the node-core package.

packages/neo-one-csharp-core/src/TrimmedBlock.ts
```diff
@@ -91,7 +94,13 @@ export class TrimmedBlock extends BlockBase implements SerializableWire<TrimmedB
...
       consensusData: this.consensusData,
-      transactions: this.hashes.slice(1).map(cache.get),
+      transactions: await Promise.all(
+        this.hashes.slice(1).map(async (hash) => {
+          const state = await storage.get(hash);
+
+          return state.transaction;
+        }),
+      ),
     });
```
spencercorwin, line 104: You can do this?! My life is changed forever

Issue #1882, opened 2019-12-02
[Neo 3.0] Update / Remove RPC commands


danwbyrne:
https://github.com/Tommo-L/NEO3-Development-Guide#rpc

20+ commands have been removed from the RPC and several others have been reworked based on NEO model changes and will need to be updated.

Some of the reworks *may* require their own issues but for now we can lump them all in together.

- [ ] remove old unnecessary commands
- [ ] update `getblockheader`
- [ ] update `getrawmempool`
- [ ] update `getblock`
- [ ] update `getblockheader`
- [ ] update `getrawtransaction`
- [ ] update `getversion`
- [ ] update `getcontractstate`

danwbyrne:
#151

danwbyrne:
https://github.com/Tommo-L/NEO3-Development-Guide/tree/master/en/RPC/api

do these wallet APIs just go into https://github.com/neo-one-suite/neo-one/blob/6b4ce5538cdaa79c0317cc89e988b807ad35411f/packages/neo-one-node-rpc-handler/src/createHandler.ts#L55 with the other RPC_METHODS?

spencercorwin:
Duplicate of #2176

PR #2214, opened 2020-11-03
fix(rpc): fix rpc getnep5transfers method for received

spencercorwin:
### Description of the Change

Fixes RPC method `getnep5transfers`, which was returning the wrong data for `received`.

Also gets closer to the C# input and return formats. Addresses in transfer objects should be base58 addresses instead of hex addresses. And inputs should take a scriptHash or base58 address.

This PR also includes a limit for `getAllStorage` RPC method for Native contracts.

### Test Plan

Was tested manually with Postman to compare against Neo TestNet node results. Will also be tested more thoroughly with unit tests later.

### Alternate Designs

None.

### Benefits

Closer to Neo3. Fixes bug.

### Possible Drawbacks

None.

### Applicable Issues

#2009 #1882

### DIFF
--- a/packages/neo-one-node-rpc-handler/src/createHandler.ts
+++ b/packages/neo-one-node-rpc-handler/src/createHandler.ts
@@ -4,6 +4,7 @@ import {
   crypto,
   JSONHelper,
   RelayTransactionResultJSON,
+  scriptHashToAddress,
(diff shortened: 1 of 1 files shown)


danwbyrne (approved):
LGTM

Issue #2212, opened 2020-10-26
Implement `getapplicationlog` RPC method and related JSONRPCClient methods/tests/types


spencercorwin:
Also need to implement storage for this information

This will likely be key information for NEO Tracker down the line

Issue #2176, opened 2020-10-07
finalize RPC Handlers


danwbyrne:
- [x] getcontractstate
- [x] getrawmempool
- [x] relaytransaction
- [ ] getstorage
- [x] sendrawtransaction
- [ ] getinvocationdata
- [x] getbestblockhash
- [x] getblock
- [x] getblockcount
- [x] getblockhash
- [x] getconnectioncount
- [x] getrawtransaction
- [x] invokefunction
- [x] invokescript
- [x] submitblock
- [x] validateaddress
- [x] getpeers
- [x] ~relaystrippedtransaction~
- [x] getallstorage
- [x] ~testinvocation~
- [x] gettransactionreceipt
- [x] getvalidators
- [x] getnetworksettings
- [x] getsettings
- [ ] runconsensusnow
- [ ] updatesettings
- [ ] fastforwardoffset
- [ ] fastforwardtotime
- [ ] reset
- [ ] getneotrackerurl
- [ ] resetproject
- [x] listplugins
- [x] gettransactionheight
- [x] getversion
- [x] getapplicationlog
- [x] getnep5transfers
- [x] getnep5balances
- [x] getblockheader
- [x] closewallet
- [x] dumpprivkey
- [x] getnewaddress
- [x] getunclaimedgas
- [x] getwalletbalance
- [x] getwalletunclaimedgas
- [x] importprivkey
- [x] listaddress
- [x] openwallet
- [x] sendfrom
- [x] sendmany
- [x] sendtoaddress
- [x] UNKNOWN
- [x] INVALID

PR #2215, opened 2020-11-04
feat(applog): add applicationlog data and RPC method, plus fixes

spencercorwin:
### Description of the Change

Tons of goods stuff here. Christmas came early:
- Adds `ApplicationLog` storage and RPC method
- Removes duplicate `TriggerType` enum and puts it and some related utilities in `client-common`
- Removes some unnecessary types and fixes up some other
- Fixes up serializeJSON for BlockBase
- Minor fix to `getBlock` RPC method
- Fixes `stackItemToJSON` method (see here: https://github.com/neo-project/neo/blob/master/src/neo/VM/Helper.cs#L194)

### Test Plan

Tested with Postman and unit test in `temp.test.ts`.

### Alternate Designs

None.

### Benefits

More functionality, less bugs, closer to Neo3.

### Possible Drawbacks

None.

### Applicable Issues

#2212 
#2176
#2009


packages/neo-one-node-core/src/StackItems/StackItems.ts
```diff
@@ -153,12 +152,12 @@ export const stackItemToJSON = (item: StackItem, context?: Set<StackItem>): Stac
     case StackItemType.Buffer:
       const buffer = assertBufferStackItem(item);
 
-      return { type: 'Buffer', value: JSONHelper.writeBuffer(buffer.getBuffer()) };
+      return { type: 'Buffer', value: JSONHelper.writeBase64Buffer(buffer.getBuffer()) };
```
spencercorwin, line 155: Buffer and ByteString are both written as base64 strings, not hex strings: https://github.com/neo-project/neo/blob/master/src/neo/VM/Helper.cs#L194

packages/neo-one-node-core/src/StackItems/StackItems.ts
```diff
@@ -187,6 +186,6 @@ export const stackItemToJSON = (item: StackItem, context?: Set<StackItem>): Stac
       return { type: 'Pointer', value: pointer.position };
 
     default:
-      throw new InvalidFormatError(`didn't expect stack item of type: ${type} on return`);
+      return { type: 'Any', value: undefined };
```
spencercorwin, line 189: If the type was `'Any'` we would throw, which was not correct and caused incorrect return values when JSON serializing some StackItems. It should actually return a JSON object with type equal to the type and value empty (ie. undefined). `'Any'` is the only type remaining if it makes it to `default` here, so we put `'Any'`.

spencercorwin: See here: https://github.com/neo-project/neo/blob/master/src/neo/VM/Helper.cs#L230

Issue #2155, opened 2020-08-30
Finish final node updates


spencercorwin:
These are the packages that will likely be updated in the node very last, after every other node package is 100% done. These require little to no changes/updates.

Packages:
- `node`
	- Create the `FullNode` class
- `node-bin`
- `node-http-rpc`
	- Just sets up the RPC server. Very basic
- `node-tools`
	- Syncs gcloud bucket local file system for node data

PR #2217, opened 2020-11-18
feat(transactions): fix all transaction stuff, rpc calls, more

spencercorwin:
### Description of the Change

Updates many things:

- `createGetHashData` function
- How `message`/`messageMagic` is passed around to the appropriate places where it's needed (like in ConsensusPayload, FeelessTransactionModel, TransactionModel, Block, BlockBase, TrimmedBlock, BlockHeader)
- Makes fixes to RPC server to match Neo C# Node implementation
- Updates types and methods in JSONRPCClient and RPC methods
- Adds a test to show how transactions should be signed (`sendrawtransaction.test.ts`)

### Test Plan

See `sendrawtransaction.test.ts` and `temp.test.ts` in `neo-one-client-core/src/__tests__/provider`.

Also tested manually with Postman.

### Alternate Designs

None.

### Benefits

Less bugs, closer to Neo3 Preview 3.

### Possible Drawbacks

None.

### Applicable Issues

#2176 
#2009 
#1882 
#2155

### DIFF
--- a/packages/neo-one-client-common/src/models/transaction/FeelessTransactionModel.ts
+++ b/packages/neo-one-client-common/src/models/transaction/FeelessTransactionModel.ts
@@ -12,6 +12,7 @@ import { AttributeModel } from './attribute';
 export interface TransactionConsensusOptions {
   readonly nonce: number;
   readonly validUntilBlock: number;
+  readonly messageMagic: number;
 }
 
 export interface TransactionFeesAdd {
@@ -26,12 +27,13 @@ export interface FeelessTransactionModelAdd<
 > {
   readonly version?: number;
   readonly nonce?: number;
-  readonly validUntilBlock?: number;
+  readonly validUntilBlock: number;
   readonly attributes?: readonly TAttribute[];
   readonly signers?: readonly TSigner[];
   readonly script: Buffer;
   readonly witnesses?: readonly TWitness[];
   readonly hash?: UInt256;
+  readonly messageMagic: number;
 }
 
 export const MAX_TRANSACTION_ATTRIBUTES = 16;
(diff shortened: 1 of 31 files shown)


packages/neo-one-client-common/src/models/transaction/FeelessTransactionModel.ts
```diff
@@ -84,7 +82,7 @@ export class FeelessTransactionModel<
     this.validUntilBlock = validUntilBlock;
     this.script = script;
     const hashIn = hash;
-    this.hashInternal = hashIn === undefined ? utils.lazy(() => crypto.hash256(this.message)) : () => hashIn;
+    this.hashInternal = hashIn === undefined ? (magic) => crypto.hash256(this.getMessage(magic)) : () => hashIn;
```
danwbyrne, line 85: I think this is way worse for how stuff will flow now, especially when looking at the `clone` method.

We should either have a `this.magic` property that gets set in the constructor or from the deserializeWireOptions context, or have a utils function that takes in a serializable. This constant passing through of magic is no bueno imo.

danwbyrne (changes requested):
I only made 1 comment but I think it stands for the other classes as well.

packages/neo-one-node-core/src/payload/UnsignedConsensusPayload.ts
```diff
@@ -86,8 +86,8 @@ export class UnsignedConsensusPayload implements SerializableWire {
 
   public get consensusMessage() {
     if (this.consensusMessageInternal === undefined) {
-      // TODO: Implement an error here;
-      throw new Error('need a way to get consensusMessage');
+      // TODO: Implement an error here
```
danwbyrne, line 89: Woah

danwbyrne (approved):
LGTM

Issue #2240, opened 2020-12-07
Reimplement retrieving `Log` info from the node


danwbyrne:
title, we already added support for handling notifications but we didn't add `Log`s back.

danwbyrne:
will implement the ability to get the logs in an upcoming PR but for the implementation of it we should start at `@neo-one/smart-contract-compiler/src/__data__/helpers/extractor.ts`

PR #2263, opened 2020-12-15
feat(misc): various fixups, VM Logging

danwbyrne:
- Adds `Log` storage to the our Neo fork and publish Neo-N1-Fork v3.0.1-preview3
- Adds the accompanying support to `@neo-one/node-vm` for retrieving these logs.
- Adds support for retrieving the logs on client side.

We also:
- fixup `provider.call(...)`
- removes `invokeTransaction` from the client and node since it was unused.
- change `testInvoke` to take just a script instead of a fully formed transaction since the transaction wasn't used anyway.
- apply prettier formatting to `@neo-one/node-vm` `.cs` files since they weren't applied before.
- fixup ByteStringStackItem value conversion when passing the JSON version to the client.

TODO: #2240


spencercorwin (approved):
LGTM

Issue #2276, opened 2020-12-29
Misc updates for preview4 changes


danwbyrne:
These didn't have a specific place in my other issues so I will group them here and move them as needed.

- [x] (1) remove `testMode` from applicationEngine https://github.com/neo-project/neo/pull/1819
- [x] (2) update `isMultisigContract` method https://github.com/neo-project/neo/pull/1890
- [x] (3) new `CheckReturnType` enum look into its use https://github.com/neo-project/neo/pull/1933
- [x] (4) Fix witness script limits https://github.com/neo-project/neo/pull/1958
- [x] (5) investigate this 'fix native serialize' change https://github.com/neo-project/neo/pull/1988
- [x] (6) update MaxValidUntilBlock https://github.com/neo-project/neo/pull/2042
- [x] (7) check this https://github.com/neo-project/neo/pull/2106/files
- [x] (8) add 'data' property to Transfer https://github.com/neo-project/neo/pull/2112/files
- [x] (9) call flag changes https://github.com/neo-project/neo/pull/2121 presumably done but we didn't really use it besides none before
- [x] (10) scriptbuilder https://github.com/neo-project/neo/pull/2145
- [x] (11) more transaction cost changes https://github.com/neo-project/neo/pull/2149/files
- [x] (12) 'nativeActivations' to protocol settings https://github.com/neo-project/neo/pull/2141

danwbyrne:
closed with #2290 #2284

Issue #2273, opened 2020-12-29
Update NativeContract models for Preview4 changes


danwbyrne:
- [x] (1) Optimize `BlockAccount` in native Policy contract https://github.com/neo-project/neo/pull/1843
- [x] (2) update native contracts for incentive changes (fairly substantial) https://github.com/neo-project/neo/pull/1845 https://github.com/neo-project/neo/pull/1847
- [x] (3) another neo token update https://github.com/neo-project/neo/pull/1917
- [x] (4) add committee cache changes https://github.com/neo-project/neo/pull/1920
- [x] (5) investigate this "fix candidate" change https://github.com/neo-project/neo/pull/1935
- [x] (6) Policy contract changes https://github.com/neo-project/neo/pull/1938
- [x] (7) 'Designate' contract https://github.com/neo-project/neo/pull/1950
- [x] (8) 'fix verifywitnesses()' of native contracts (we might not need this?) https://github.com/neo-project/neo/pull/1998
- [x] (9) other voter incentives https://github.com/neo-project/neo/pull/1848
- [x] (10) oracle / designate contract changes https://github.com/neo-project/neo/pull/2007
- [x] (11) more reward changes https://github.com/neo-project/neo/pull/2049
- [x] (12) oracle contract https://github.com/neo-project/neo/pull/2036
- [x] (13) NEP17 https://github.com/neo-project/neo/pull/2024/
- [x] (14) management contract https://github.com/neo-project/neo/pull/2119
- [x] (15) adjustable opcode pricing https://github.com/neo-project/neo/pull/2045
- [x] (16) asset symbols https://github.com/neo-project/neo/pull/2136/files
- [x] (17) minimum deployment fee https://github.com/neo-project/neo/pull/2140

danwbyrne:
closed with #2290 #2284

PR #2279, opened 2020-12-29
feat(vm): remove testMode, fixup vm, various quick fixes

danwbyrne:
fixup issues 1,4,6,9 and 12 from #2276 

removes testmode and fixes up some of the various changes made in the VM since preview3 that we will need.

Contract storage will be re-implemented with the `management` contract in #2273 

Will continue to use my local neo project until I'm confident there is nothing else that needs to be changed before releasing the preview4 fork.


spencercorwin:
Should this be on the preview 4 branch? @danwbyrne

Issue #2278, opened 2020-12-29
Update blockchain for preview4 changes


danwbyrne:
- [x] (1) investigate this change for calculating verificationCost https://github.com/neo-project/neo/pull/1931
- [x] (2) onPersist / postPersist + trigger changes https://github.com/neo-project/neo/pull/2022
- [x] (3) asynchronous oracle blockchain changes https://github.com/neo-project/neo/pull/1738
- [x] (4) check these changes in blockchain for oracle https://github.com/neo-project/neo/pull/1884

danwbyrne:
closed with #2284  #2290

PR #2284, opened 2021-01-05
Preview4 blockchain changes

danwbyrne:
commit 3 adds:
implement changes 1 and 2 from #2278

Fixes an old issue in the neo-one/node-vm storage implementation. TL;DR the C# version has the table property it can use for indexing but since we don't have that in JS we "cheat" by just appending the table to the key and seeking off of that. This was an issue when deserializing the key for use in the new GasPerBlockRecords storage lookup in the NEO Native contract and was probably an issue before but since we didn't use the key AFTER seeking like this before it didn't come up.

commits 1-2 add:
fixup issues 1,4,6,9 and 12 from #2276

removes testmode and fixes up some of the various changes made in the VM since preview3 that we will need.

Contract storage will be re-implemented with the management contract in #2273

Will continue to use my local neo project until I'm confident there is nothing else that needs to be changed before releasing the preview4 fork.


packages/neo-one-node-vm/lib/Dispatcher.csproj
```diff
@@ -14,8 +14,12 @@
...
     <PackageReference Include="RocksDbNative" Version="6.2.2" />
     <PackageReference Include="RocksDbSharp" Version="6.2.2" />
-    <PackageReference Include="Neo.VM" Version="3.0.0-preview3" />
-    <PackageReference Include="Neo-N1-Fork" Version="3.0.1-preview3" />
+    <PackageReference Include="Neo.VM" Version="3.0.0-preview4" />
+    <!-- <PackageReference Include="Neo-N1-Fork" Version="3.0.1-preview3" /> -->
+  </ItemGroup>
+
+  <ItemGroup>
+    <ProjectReference Include="/Users/danielbyrne/neo/src/neo/neo.csproj" />
```
spencercorwin, line 22: Undo

Issue #2275, opened 2020-12-29
Update consensus models for preview4 changes


danwbyrne:
- [x] Optimize `LastSeenMessage` in ConsensusContext https://github.com/neo-project/neo/pull/1838
- [x] optimize nextConsensus https://github.com/neo-project/neo/pull/2015

danwbyrne:
closed with #2290

Issue #2270, opened 2020-12-29
implement neo commit #2044


danwbyrne:
final task of #2269 will be tracked here

PR #2290, opened 2021-01-13
feat(misc): final preview4 changes

danwbyrne:
Final PR for Preview4 updates,

Fixes the remaining issues of #2276  as well as #2275 and #2270 

Also fixes several lint errors that were previously ignored because of my ESLint install.

As a final fix we update `nep5` naming conventions to `nep17`


packages/neo-one-website/docs/4-contributing/1-codebase-overview.md
```diff
@@ -158,7 +158,7 @@ Creates the environment for compiling smart contracts, which is then used in the
 
 ### neo-one-smart-contract-lib
 
-Defines NEP5 tokens and ICOs.
+Defines NEP17 tokens and ICOs.
```
spencercorwin, line 161: Actually revert this since NEP5 will still be NEP5 for Neo2 docs, which will still be produced from the `docs` folder. I've made this change to the new Neo3 docs folder on my branch.

PR #2348, opened 2021-03-10
fix(compiler, doc): mixin syntax fix and template doc.

ndhuutai:
### Requirements

Template support.

### Description of the Change

ts compiler ignored contracts that use mixins syntax.
```typescript
// ...imports

export class TestToken extends NEP5Token(SmartContract){
// ...
}
```

The issue came from the following code block: 

```typescript
if (clause === undefined || !ts.isHeritageClause(clause) || !heritage.isExtends(clause)) {
  return acc;
}
```
The `NEP5Token(SmartContract)` is neither a HeritageClause nor an Extends clause. So a separate `if` seems necessary for the specific case of mixins.

### Test Plan

1. symlink `neo-one-ts-utils`
2. use mixins syntax like above
3. run compile or build to see results

### Alternate Designs

N/A

### Benefits

Able to compile contracts with mixins.

### Possible Drawbacks

Might affect other syntaxes.

### Applicable Issues

N/A

### DIFF
--- a/packages/neo-one-ts-utils/src/class_.ts
+++ b/packages/neo-one-ts-utils/src/class_.ts
@@ -272,9 +272,10 @@ function getImplementorsWorker(
         return acc;
       }
 
-      let derived: ts.ClassDeclaration | ts.InterfaceDeclaration | undefined = node_.getFirstAncestorByKind<
-        ts.ClassDeclaration
-      >(clause, ts.SyntaxKind.ClassDeclaration);
+      let derived:
+        | ts.ClassDeclaration
+        | ts.InterfaceDeclaration
+        | undefined = node_.getFirstAncestorByKind<ts.ClassDeclaration>(clause, ts.SyntaxKind.ClassDeclaration);
       if (derived === undefined) {
         derived = node_.getFirstAncestorByKindOrThrow<ts.InterfaceDeclaration>(
           clause,
@@ -314,7 +315,13 @@ function getExtendorsWorker(
       }
 
       const clause = node_.getParent(parent) as ts.Node | undefined;
-      if (clause === undefined || !ts.isHeritageClause(clause) || !heritage.isExtends(clause)) {
+      // if (clause === undefined || !ts.isHeritageClause(clause) || !heritage.isExtends(clause)) {
+      //   return acc;
+      // }
+
+      if (clause && ts.isExpressionWithTypeArguments(clause)) {
+        // do nothing and continue.
+      } else if (clause === undefined || !ts.isHeritageClause(clause) || !heritage.isExtends(clause)) {
         return acc;
       }
 
--- a/packages/neo-one-ts-utils/src/class_.ts
+++ b/packages/neo-one-ts-utils/src/class_.ts
@@ -315,13 +315,11 @@ function getExtendorsWorker(
       }
 
       const clause = node_.getParent(parent) as ts.Node | undefined;
-      // if (clause === undefined || !ts.isHeritageClause(clause) || !heritage.isExtends(clause)) {
-      //   return acc;
-      // }
 
-      if (clause && ts.isExpressionWithTypeArguments(clause)) {
-        // do nothing and continue.
-      } else if (clause === undefined || !ts.isHeritageClause(clause) || !heritage.isExtends(clause)) {
+      if (
+        (clause === undefined || !ts.isHeritageClause(clause) || !heritage.isExtends(clause)) &&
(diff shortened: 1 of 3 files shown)


packages/neo-one-ts-utils/src/class_.ts
```diff
@@ -314,7 +315,10 @@ function getExtendorsWorker(
       }
 
       const clause = node_.getParent(parent) as ts.Node | undefined;
-      if (clause === undefined || !ts.isHeritageClause(clause) || !heritage.isExtends(clause)) {
+
+      if (clause && ts.isExpressionWithTypeArguments(clause)) {
+        // do nothing and continue.
+      } else if (clause === undefined || !ts.isHeritageClause(clause) || !heritage.isExtends(clause)) {
```
spencercorwin, line 321: This is a strange pattern. I'd prefer you combined it into one `if` statement like so:
```
if (clause === undefined || !ts.isHeritageClause(clause) || !heritage.isExtends(clause) && !(clause === undefined && ts.isExpressionWithTypeArguments(clause))) {
   return acc;
}
```

Or whatever makes sense with that logic

packages/neo-one-website/docs/0-installation/5-contract-templates.md
```diff
@@ -0,0 +1,106 @@
...
+
+> A mixin is an abstract subclass; i.e. a subclass definition that may be applied to different superclasses to create a related family of modified classes.
+>
+> -Gilad Bracha and William Cook, [Mixin-based Inheritance](http://www.bracha.org/oopsla90.pdf)
+
+::: warning
+
+Tip
+
+It is encouraged that you spend time to explore how we implemented NEP5Token as well as other contract templates
```
spencercorwin, line 62: "We encourage you to look at the `NEP5Token` mixin to understand how we are implementing a NEP5 token. Go [here](https://github.com/neo-one-suite/neo-one/blob/master-2.x/packages/neo-one-smart-contract-lib/src/NEP5Token.ts) to see the source code for the `NEP5Token` mixin."

packages/neo-one-website/docs/0-installation/5-contract-templates.md
```diff
@@ -0,0 +1,106 @@
...
+
+## Available Templates
+
+::: warning
+
+Note
+
+`NEP5Token` is being updated to `NEP17Token`. Stay tuned!
+
+:::
```
spencercorwin, line 101: I say we remove this to avoid confusion. The docs on the current website are for Neo2 only and NEP17 is for Neo3 only I believe.

If I'm wrong about that then disregard.

packages/neo-one-website/docs/0-installation/5-contract-templates.md
```diff
@@ -0,0 +1,106 @@
...
+}
+```
+
+If you find the syntax `NEP5Token(SmartContract)` strange, checkout this page on [Mixins](https://www.typescriptlang.org/docs/handbook/mixins.html) from the TypeScript Handbook for more details.
+
+Addtionally, you can also checkout this blog post - ["Real" Mixins with JavaScript Classes](https://justinfagnani.com/2015/12/21/real-mixins-with-javascript-classes/) - to see how mixins work under the hood.
+
+> A mixin is an abstract subclass; i.e. a subclass definition that may be applied to different superclasses to create a related family of modified classes.
+>
+> -Gilad Bracha and William Cook, [Mixin-based Inheritance](http://www.bracha.org/oopsla90.pdf)
```
spencercorwin, line 56: Remove the quote and PDF link and write this same sentence in a slightly more user friendly way. Or just use the sentence as it is if you think this is the clearest way to explain a mixin.

spencercorwin (changes requested):

packages/neo-one-website/docs/0-installation/5-contract-templates.md
```diff
@@ -0,0 +1,141 @@
...
+    }
+    this.owner = owner;
+    this.issue(owner, amount);
+  }
+}
+```
+
+`Redtoken.ts`
+
+`RedToken` inherits methods and properties that adheres to the NEP5Token standard because of the mixin. It also inherits everything defined in `SimpleToken`.
```
spencercorwin, line 118: "...the NEP5 token standard..."

spencercorwin (changes requested):

spencercorwin:
Also need to look at linter errors in CI and fix those

spencercorwin (approved):

